PR #6, opened 2017-02-01
Update links and descriptions in pom.xml

coudot:
I made some changes in pom.xml to be up to date with project migration (github and new server).

As now documented in contribution process (http://lsc-project.org/development#process), I submit it as pull request. Without any feedback within 3 days, I will apply it. But if it is ok for you, please comment as soon as possible.

I will also apply these modifications to 2.1 branch.


coudot:
Well I was not sure if we should have an issue and a pull request, as pull request can also be associated to a milestone, but not directly to an issue. Indeed each commit can refer to an issue, and will also be associated to the pull request.

Do we decide to always create an issue, then create a pull request in which all commits reference the issue?

rouazana:
How do you connect a commit to its PR? I don't think it feasible, so creating issues and referencing them in commit seems still useful.

rouazana:
When you merge, you should not squash all commits into one, you loose your (fine) history.

PR #7, opened 2017-02-03
V2.1 - SASL Quality of Protection

Kwoin:
With Kerberos authentication, you may have to set the property _javax.security.sasl.qop_ in order to succeed authentication, depending of the security policy

This pull request includes : 

- Added a new optional element **saslQop** into lsc-core-2.1.xsd
- Added new line to JndiService to use this new setting

I did not add generated xjc class, maven will do in each build.

(Btw, I also added IntelliJ IDEA files in .gitignore)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,11 @@
-/target/
+# Eclipse
 .classpath
 .project
 .settings/
+
+# IntelliJ IDEA
+.idea
+*.iml
+
+# bin
+/target/
--- a/src/main/java/org/lsc/jndi/JndiServices.java
+++ b/src/main/java/org/lsc/jndi/JndiServices.java
@@ -45,16 +45,27 @@
  */
 package org.lsc.jndi;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Hashtable;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
+import org.apache.commons.lang.StringUtils;
+import org.apache.directory.api.ldap.codec.api.ControlFactory;
+import org.apache.directory.api.ldap.codec.api.LdapApiService;
+import org.apache.directory.api.ldap.codec.api.LdapApiServiceFactory;
+import org.apache.directory.api.ldap.codec.controls.search.persistentSearch.PersistentSearchFactory;
+import org.apache.directory.api.ldap.extras.controls.syncrepl_impl.SyncStateValueFactory;
+import org.apache.directory.api.ldap.model.exception.LdapInvalidDnException;
+import org.apache.directory.api.ldap.model.exception.LdapURLEncodingException;
+import org.apache.directory.api.ldap.model.name.Dn;
+import org.apache.directory.api.ldap.model.name.Rdn;
+import org.apache.directory.api.ldap.model.url.LdapUrl;
+import org.lsc.Configuration;
+import org.lsc.LscDatasets;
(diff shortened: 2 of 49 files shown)


src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -101,15 +58,21 @@
...
 
+import javax.naming.*;
+import javax.naming.directory.*;
+import javax.naming.ldap.*;
+import javax.security.auth.callback.*;
+import javax.security.auth.login.LoginContext;
+import javax.security.auth.login.LoginException;
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
```
rouazana: please don't reorganize import like this, we prefer to not have .* imports

.gitignore
```diff
@@ -8,8 +8,4 @@
 *.iml
 
 # bin
-/target/
-
-# xjc generation
-src/main/java/org/lsc/configuration/
-lsc.episode
\ No newline at end of file
+/target/
```
rouazana: cannot you fix the first commit instead?

rouazana:
BTW thank you for your contribution!

src/main/java/org/lsc/jndi/JndiServices.java
Kwoin: Ah sry for that, I'll fix the IDE settings

.gitignore
Kwoin: You may prefer I squash all the commit in one ?
(Sorry, I'm not totally comfortable with pull requests yet)

rouazana: nope, having multiple commits is fine. That's just your first and your third commit that should be squashed together. You can do this easily with git rebase -i

Kwoin:
I've made required changes (squash commits and declutter imports, it seems my IDE still rearrange them a bit, I hope it does not affect your coding style too much)

coudot:
@rouazana if this is ok for you, I let you apply the PR and merge it into master

Regarding documentation, I can have a look at it. I suppose only this page needs an update: http://lsc-project.org/documentation/howto/kerberos ?

rouazana:
Sorry I won't have time to do the merge myself (it involves some non trivial rebasing I guess). So please do it, or if @Kwoin can, please provide PR a PR for master too.

coudot:
@soisik do you think you can help use to merge this PR on trunk?

soisik:
@coudot if not mistaking,you want this PL to be merged in master but not 2.1 branch, or both ?

coudot:
This PR can be merged in 2.1 directly, but it seems to need some adaptations to be merged in trunk. So I would like to have both ;)

soisik:
Done on both branches.

coudot:
Thanks a lot!

Do you think we could add a unit test on this new parameter?

coudot:
Documentation updated here : http://lsc-project.org/documentation/howto/kerberos#lsc

@Kwoin could you tell me if this is ok?

Kwoin:
@coudot 
the _**saslQop**_ element should appear after **_tlsActivated_** element, so the XSD wont complain about the sequence order.

soisik:
@coudot in documentation, you should mention that default value is "auth".

As for the junit test, do you want a full GSSAPI / kerberos auth test ? Should be possible (spring-security-kerberos has junit tests on kerberos auth). I don't see anything else to test around this feature.

coudot:
@Kwoin thanks, doc updated
@soisik maybe a simple test is to have a sample lsc.xml with this attribute set to check that the configuration is valid. Testing Kerberos is more complex, could be interesting but not needed at short term.

Issue #9, opened 2017-02-17
Doesn't run on Windows


webbird:
I'd like to check LSC for a project I am working on, but I cannot get it working on Windows Server 2008 R2 with JDK 8. I tried to run the sample from the docs.

First, in the lsc-sample.bat, there are several lines starting with a # which causes lots of syntax error messages. After removing that lines, the lsc.bat doesn't work, either. There seem to be several problems with the vars: Though JAVA_HOME is set, it is considered undefined. After replacing the JAVA_COMMAND with hardcoded java.exe path, I am still stuck as there's a huge error stack.

Has ANYONE LSC running on windows successfully?

coudot:
Hi,

which LSC version did you use?

Note that someone already has problems on Windows and reported this issue: http://tools.lsc-project.org/issues/826

Could you test what he proposed? If it works, please tell use, or even better, create a pull request with modifications.

webbird:
Thanks for your reply! I tried it on my workstation running Windows 7 x64 with latest Java 8 x64 Runtime (not SDK) and managed to get it running. Don't know why it doesn't on that Windows Server. So the only issue left are the #-lines in the lsc-sample.bat. Just replace the # with _Rem_ and that's it.

sbahloul:
I run it on Windows through Cygwin.

PR #10, opened 2017-02-17
Replace # by REM in bat scripts (references #9)

coudot:
The goal is to fix the sample on Windows


coudot:
Also applied on v2.1 branch with cherry-pick

PR #44, opened 2018-03-09
Javascript feature to search attribute values

floriancaillet:
Hello,

**Context** 
I'm using LSC to synchronize a LDAP eDirectory to a LDAP OpenDj.
The directories (source - target) haven't got the same structure.  So I had to write some javascript to build members attributes in my destination directory.
During tests, I faced performances issues with thousands source entries. 
Because my Javascript searched Dn with srcLdap.search queries and then for each dn searched a unique attribute value using  srcLdap.attribute.

**LSC changes**
I understood that if I can search directly unique attribute values using a new method  srcLdap.searchAttribute, performances will be improved.
So I developped new ScriptableJNDI services :
-searchAttribute : Performs a search with subtree scope on a given base DN with a given filter that returns attribute values.
-listAttribute : Performs a search with one level scope on a given base DN with a given filter that returns attribute values.

I think this new feature could benefit everyone.
So, I submit you the code for integration in lsc projet.
Let me now, if I need to refactor anything.

Thanks
Florian

### DIFF
--- a/src/main/java/org/lsc/jndi/ScriptableJndiServices.java
+++ b/src/main/java/org/lsc/jndi/ScriptableJndiServices.java
@@ -48,6 +48,8 @@ package org.lsc.jndi;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import org.lsc.LscDatasets;
 
 import javax.naming.NamingException;
 import javax.naming.directory.SearchControls;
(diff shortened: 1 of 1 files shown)


floriancaillet:
I've just made a new commit and formatting issues are away.
Best regards.

src/main/java/org/lsc/jndi/ScriptableJndiServices.java
```diff
@@ -126,6 +128,59 @@ public final JndiServices getJndiServices() {
...
+		List<String> resAttributes = new ArrayList<String>();
+
+		List<String> attrsNames = new ArrayList<String>();
+		attrsNames.add(attribute);
+
+		// Searching attributes values
+		Map<String, LscDatasets> res = jndiServices.getAttrsList(base, //
+			filter, //
+			scope, //
+			attrsNames //
```
rouazana: can you fix these strange comments lines please?

src/main/java/org/lsc/jndi/ScriptableJndiServices.java
```diff
@@ -126,6 +128,59 @@ public final JndiServices getJndiServices() {
...
+		);	
+
+		for (Map.Entry<String, LscDatasets> entry : res.entrySet()) {
+	    	resAttributes.add(entry.getValue().getStringValueAttribute(attribute));
```
rouazana: missing indentation

floriancaillet:
I squash commits. It's ok thanks.

coudot:
We could maybe backport this in 2.1 version.

I think we also need a documentation page to show all possibilities of JNDI services.

soisik:
Hi,
Actually it's already backported on 2.1.
I think too a page with a few examples could be useful. For now, there is a link to javadoc from home page; https://lsc-project.org/documentation/latest/start?s[]=scriptablejndiservices. The javadoc contains some indications on how to use it and how methods can be called.

PR #48, opened 2018-03-16
Hotfix/asyncmode decoder exception

floriancaillet:
Hello, 

I found a bug working in asynchronous mode with a Novell Edirectory source.

Error is Unexpected DecoderException: The length of controls must not be null.
This bug is known on apache.directory.api (https://issues.apache.org/jira/browse/DIRAPI-295) and has been fixed in 1.0.0 version of org.apache.directory.api api-all.

So I decide to update api-all from 1.0.0-M22 to 1.0.0.
There is an impact in JndiServices and SyncReplSourceService classes. And the project has to run in JAVA 1.8.

Do you think this fix could be integrated to new lsc version ?

Thanks,
Florian


rouazana:
It is possible to have a clean changelog? Here I have no idea of what it to review.

coudot:
there is a conflict with old issues

rouazana:
it seems that the base branch of this PR does not match the bas branch of the fix

floriancaillet:
I will make an new PR whith de correct branch.

PR #49, opened 2018-03-19
Fix on asynchronous mode with a Novell Edirectory source

floriancaillet:
Hello,

I found a bug working in asynchronous mode with a Novell Edirectory source.

Error is Unexpected DecoderException: The length of controls must not be null.
This bug is known on apache.directory.api (https://issues.apache.org/jira/browse/DIRAPI-295) and has been fixed in 1.0.0 version of org.apache.directory.api api-all.

So I decide to update api-all from 1.0.0-M22 to 1.0.0.
There is an impact in JndiServices and SyncReplSourceService classes. And the project has to run in JAVA 1.8.

Do you think this fix could be integrated to new lsc version ?

Thanks,
Florian

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -630,6 +630,17 @@
 			</exclusions>
 		</dependency>
 		<dependency>
+			<groupId>xmlrpc</groupId>
+			<artifactId>xmlrpc</artifactId>
+			<version>3.0a1</version>
+		</dependency>
+		<dependency>
+			<groupId>xmlrpc</groupId>
+			<artifactId>xmlrpc-common</artifactId>
+			<version>3.0</version>
+		</dependency>
+		
+		<dependency>
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
 			<version>1.4</version>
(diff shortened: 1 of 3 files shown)


floriancaillet:
Without  these 2 xmlrpc dependencies, I can't build the project with maven (Which Maven repo did you use ?)
I don't test with JAVA 1.7, but it probably works. For sure it doesn't work in JAVA 1.6.

soisik:
I'm using maven central. Could you send me your error log and tell me the maven & java version you use?

floriancaillet:
Now I can build the project without these 2 xmlrpc dependencies changing my enterprise Maven Settings.
I 've just commited a new version of the pom.xml.

soisik:
@coudot @rouazana please note that with this PR, LSC will not be supporting java 1.6 anymore (tests fails on java 1.6 while it works on current v2.1). Page https://lsc-project.org/documentation/latest/requirements will have to be updated. Maybe this should be merged into master, isn't v2.1 just for minor release ? Please advice. Otherwise I'm OK with this PR.

rouazana:
I agree, for me we should not break compatibility in a minor release (even if nobody should use java 1.6).

coudot:
Ok, in this case this PR should only be for LSC 2.2, but I hope we can release this major version in the next months.

coudot:
See #106

PR #57, opened 2018-05-15
Fixing typo

joaoappereira:


coudot:
Thanks for the PR. We will merge it and apply it also to 2.1 version.

src/main/java/org/lsc/AbstractSynchronize.java
```diff
@@ -656,7 +656,7 @@ public void run() {
             } else if (task.getSourceService() instanceof IAsynchronousService) {
                 aService = (IAsynchronousService) task.getSourceService();
             } else {
-                LOGGER.error("LSC should never reach this point ! Please consider debugging the code because we are trying to launch an asynchronous sync without any asynchronous servoice !"); 
+                LOGGER.error("LSC should never reach this point ! Please consider debugging the code because we are trying to launch an asynchronous sync without any asynchronous service !");
```
rouazana: While you are at it, you can remove spaces before each exclamation point.

PR #89, opened 2019-09-13
Fixed non existing function called in sample lsc xml file

soisik:
LSC sample file installed under /etc/lsc contains a call to a non existing function.


soisik:
Backported to 2.1

PR #91, opened 2019-09-17
sort xsd files list

remk:
For the moment the list of xsd files is loaded in the reverse order of  their hash.
so `lsc-openpaas-plugin-1.0.xsd` will be loaded after `schemas/lsc-core-2.1.xsd`
but  `lsc-james-plugin-1.0.xsd` will be loaded before and the call to ` schemaFactory.newSchema` will fail with a `SAXException` as the definitions in lsc core are required in the plugin schema.

The hash is not easily determinable and is an implementation details which is not desirable.
For example `lsc-james0-plugin-1.0.xsd`  is loaded in the correct order but is not `lsc-james-plugin-1.0.xsd`.

There is also a risk to change the order of the hashes of the core and the plugin by incrementing the version of the files.

This proposal aims to always load the core xsd first then the others in alphabetical order.


soisik:
This PR was backported to branch v2.1

rouazana:
> This PR was backported to branch v2.1

Thank you very much!

PR #92, opened 2019-09-20
Inline env variables in xml configuration

remk:
The goal of this PR is to provide a way to use environment variables in the XML configuration file.
The supported format of the environment variable in the file is `${ENVIRONMENT_VARIABLE}`

The replacement is done before the validation of the file with the xml schema in order to validate the actual configuration.


src/test/java/org/lsc/configuration/ConfigurationLoaderTest.java
```diff
@@ -10,38 +10,126 @@
...
+		assertEquals("ldap://" + LDAP_CONNECTION_HOST_ENV_VALUE + ":" + LDAP_CONNECTION_PORT_ENV_VALUE
+				+ "/dc=lsc-project,dc=org", connectionWithEnvInlined.url);
+		assertEquals(LDAP_CONNECTION_PASSWORD_ENV_VALUE, connectionWithEnvInlined.password);
+		assertEquals(LDAP_CONNECTION_USERNAME_PART_1_ENV_VALUE + LDAP_CONNECTION_USERNAME_PART_2_ENV_VALUE,
+				connectionWithEnvInlined.username);
+
+	}
+	
+	@Test
+	public void valueFromEnvVariableShoulbBeEscapedWhenLoadingConfigurationWithEnvVariables() throws FileNotFoundException, LscConfigurationException {
```
rouazana: s/Shoulb/Should

src/main/java/org/lsc/configuration/JaxbXmlConfigurationHelper.java
```diff
@@ -113,56 +119,102 @@ public JaxbXmlConfigurationHelper() throws LscConfigurationException {
...
+					LOGGER.error("Error during closing the configuration file " + filename, e);
+				}
+			}
+		}
+	}
+	
+	private String inlineEnvForLine(Map<String, String> env, String line) {
+		String resultString = line;
+		
+		for(String envVariableKey : env.keySet()) {
```
rouazana: can you quickly check if doing this for every env variable does not slow down parsing xml files too much? If this is the case we could add a little optimization checking before if the line contains `${`

remk: It seems to not have much of an inpact on the 232 lines xml test file.
i have an average time for 100 runs of loading the file of : 
- 178 ms for the file without environment variable placeholder and no environment variable provided.
- 180 ms for the file with environment variable placeholder and 110 environment variables.

coudot:
Very interesting, could be used to get current time and adapt search filter

PR #94, opened 2019-09-27
Make LSC compatible with JDK 11, remove compatibility with JDK 5, 6, 7

rouazana:
The goal of this pull request is to make LSC (future 2.2 version) compatible with only supported Java version: 8 et 11.


rouazana:
next step would be to update the Maven dependencies used by LSC.

Issue #98, opened 2019-10-28
Add pivot value in error message when no source object found


coudot:
We have sometimes these errors:
```
oct. 28 14:25:17 - ERROR - Synchronization aborted because no source object has been found !
```

Which occurs when an entry has been listed in getAllFilter but is not found with getOneFilter. It would be great to set the pivot value in the error message so we can quickly know which was the entry in error.

coudot:
Thanks a lot @soisik, I tested this today, it works well!

PR #99, opened 2019-10-30
Add pivots in source object not found error message #98

soisik:


coudot:
Wonderful, please merge when you can!

Issue #18, opened 2017-05-22
The lsc user created by the debian package should ever be a system user


jojo2024:
In lsc.postinst script provided by lsc 2.1.4-1 package, lsc user is created with the "useradd lsc" command.
If nss_ldap is configured to map ldap user, the lsc user can obtain an uidNumber in the "ldap range" instead of the "system range".
Adding the "--system" option can avoid this problem.

kasimon:
Also keep in mind that this fix would only affect new installations, not upgrades. I would recommend adding a note to the release notes or README.Debian that it may be better to purge and reinstall the package to fix the lsc user. I don't think there's a safe way to retroactively "systemize" an already created user account.

coudot:
Merged in 2.1 and applied in master with cherry-pick

PR #113, opened 2020-02-10
Create LSC group and user as system

coudot:
References #18

### DIFF
--- a/src/install/debian/lsc.postinst
+++ b/src/install/debian/lsc.postinst
@@ -4,8 +4,8 @@ lsc_logdir=/var/log/lsc
 lsc_user=lsc
 lsc_group=lsc
 
-getent group ${lsc_group} > /dev/null 2>&1 || groupadd -r ${lsc_group}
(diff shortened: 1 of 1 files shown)


src/install/debian/lsc.postinst
```diff
@@ -4,7 +4,8 @@ lsc_logdir=/var/log/lsc
 lsc_user=lsc
 lsc_group=lsc
 
-id ${lsc_user} &>/dev/null || useradd ${lsc_user}
+getent group ${lsc_group} > /dev/null 2>&1 || groupadd -r ${lsc_group}
+getent passwd ${lsc_user} > /dev/null 2>&1 || useradd -r -g ${lsc_group} -d /etc/lsc ${lsc_user}
```
rouazana: a common usage for command line flags is "use the short form when you type in you terminal, use the long form when you write a script". It allows a better readability. Here I think in particular `--system` is really more clear than `-r`

PR #69, opened 2018-08-14
Fix NPE when job has two tasks with database destinations and sync op…

dean-radish:
[lsc-test.zip](https://github.com/lsc-project/lsc/files/2287732/lsc-test.zip)
…tions

Null pointer thrown in job containing multiple tasks with database destinations and sync options. 

java.lang.NullPointerException: null
        at org.lsc.beans.syncoptions.PropertiesBasedSyncOptions.getDefaultValuedAttributeNames(PropertiesBasedSyncOptions.java:129) ~[bin/:na]
        at org.lsc.beans.BeanComparator.getWriteAttributes(BeanComparator.java:446) ~[bin/:na]
        at org.lsc.beans.BeanComparator.getUpdatedObject(BeanComparator.java:252) ~[bin/:na]
        at org.lsc.beans.BeanComparator.calculateModifications(BeanComparator.java:176) ~[bin/:na]
        at org.lsc.SynchronizeTask.run(AbstractSynchronize.java:755) [bin/:na]
        at org.lsc.SynchronizeTask.run(AbstractSynchronize.java:689) [bin/:na]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [na:1.8.0_161]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [na:1.8.0_161]
        at java.lang.Thread.run(Unknown Source) [na:1.8.0_161]

Test case attached

PR #115, opened 2020-02-14
Enhancement for #69 Fix Null pointer thrown in job containing multiple tasks with database destinations and sync options.

rouazana:
See #69


coudot:
Applied to master with cherry-pick

PR #116, opened 2020-04-10
Allow ContextDn with empty string

cheshenglin:
The origin version would throw a null pointer exception when we setting the <BaseDn> as an empty string in configuration XML. Here I modify the getContextDn() method and avoid invoking contextDn.toString() directly to fix this problem.

### DIFF
--- a/src/main/java/org/lsc/jndi/JndiServices.java
+++ b/src/main/java/org/lsc/jndi/JndiServices.java
@@ -557,9 +557,9 @@ public final class JndiServices {
 		try {
 			sc.setSearchScope(scope);
 			String rewrittenBase = null;
-			if (contextDn != null && searchBase.toLowerCase().endsWith(contextDn.toString().toLowerCase())) {
-				if (!searchBase.equalsIgnoreCase(contextDn.toString())) {
-				rewrittenBase = searchBase.substring(0, searchBase.toLowerCase().lastIndexOf(contextDn.toString().toLowerCase()) - 1);
+			if (contextDn != null && searchBase.toLowerCase().endsWith(getContextDn().toLowerCase())) {
+				if (!searchBase.equalsIgnoreCase(getContextDn())) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -557,9 +557,9 @@ private SearchResult doGetEntry(final String base, final String filter,
 		try {
 			sc.setSearchScope(scope);
 			String rewrittenBase = null;
-			if (contextDn != null && searchBase.toLowerCase().endsWith(contextDn.toString().toLowerCase())) {
-				if (!searchBase.equalsIgnoreCase(contextDn.toString())) {
-				rewrittenBase = searchBase.substring(0, searchBase.toLowerCase().lastIndexOf(contextDn.toString().toLowerCase()) - 1);
+			if (contextDn != null && searchBase.toLowerCase().endsWith(getContextDn().toLowerCase())) {
```
rouazana, line 560: here there is first a check that contextDn is not null, so the change on this line seems useless

cheshenglin: Yes, you are right! Thanks.

soisik (approved):
@rouazana is this PR ok for you too ?

rouazana:
> @rouazana is this PR ok for you too ?

It would probably be simpler and safer to make it final (if possible) and to initialize it at `"" ` instead of null here: https://github.com/lsc-project/lsc/blob/v2.1/src/main/java/org/lsc/jndi/JndiServices.java#L224

soisik:
> > @rouazana is this PR ok for you too ?
> 
> It would probably be simpler and safer to make it final (if possible) and to initialize it at `"" ` instead of null here: https://github.com/lsc-project/lsc/blob/v2.1/src/main/java/org/lsc/jndi/JndiServices.java#L224

contextDn cannot be made final as it is set when connection is initiated, not when the class is instanced.
I'll send a PR to initialize it at `"" ` as it is indeed much simple.

coudot:
Already done in #159

Issue #62, opened 2018-06-05
Configuration of multiple plugin packages fails on Windows systems


dean-radish:
Hello, configuring multiple plugin packages e.g

-DLSC.PLUGINS.PACKAGEPATH=foo.bar.plugins.configuration.file:foo.bar.plugins.configuration.ext

fails on Windows when using either a : or a ; path separator. The Unix separator ':' is hard coded in the JaxbXmlConfigurationHelper constructor, but System.getProperty("path.separator") is used later in the same class. Please make this consistent. 
Thanks
Dean

coudot:
Thanks a lot @dean-radish for the patch. It has been applied in master and 2.1 branch

PR #117, opened 2020-04-10
Fixed packageName calculation with plugin packagepath on windows.

MSSputnik:
Even on Windows the package path for the JAXBContext uses a colon. So concatenation with 'System.getProperty("path.separator")' does not work in windows systems.


MSSputnik:
Found 2nd issue with the colon (#62 and #65). Close this and open new pull request.

PR #118, opened 2020-04-11
Fix package path calculation and use always colon as separator even on Windows.

MSSputnik:
Found issues #62 and #65 where the separator was made consistent but the chosen separator 'System.getProperty("path.separator")' does not work on Windows.
The separator must be a colon. On Windows the "path.separator" is a semi colon.


MSSputnik:
Hi.
Just revert issue #65 back is not enough. 
In my issue consistent means, you need to have the same separator where ever you use the variable "LSC.PLUGINS.PACKAGEPATH" which is in 2 functions
This was the root cause of issue #62, complaining about inconsistency.

It is a windows issue only as path.separator is ":" on Linux and ";" on Windows. So this change makes no difference on Linux Systems.

And why ":" and not path.separator: 
When I use path.separator on my Windows system, the class loading fails and when checking what the loader does using procmon (Sysinternals tool) I see that the full variable content (class2;class2) is searched in the file system which is obviously wrong and will never work.
When checking the JAXBContext documentation https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/JAXBContext.html
I never see that the separator is something different than ":".

soisik (approved):

Issue #165, opened 2021-07-16
Use relax rules control when performing updates


soisik:
Some attributes (entryUUID, createTimestamp ...) cannot be updated like normal attributes, as they are protected operational attributes. 

Since Openldap supports [relax rules control](https://tools.ietf.org/html/draft-zeilenga-ldap-relax-03), we should add an option in a connection configuration to activate this control and be able to update operational attributes. 

Our use case is to keep identical entryUUID while migrating to a new LDAP directory (some apps use this attribute to link ldap entry to their local user/group), and keep historical information.

davidcoutadeur:
For information and if I remember correctly, some attributes, even if used with relax control, are not updatable

soisik:
PR merged into v2.1 and master

PR #166, opened 2021-07-16
Use relax rules control when performing updates

soisik:
See #165


coudot:
Ok to add it to 2.1.6. As the new parameter is optional it should not break existing configurations

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -903,6 +929,15 @@ private boolean doApply(final JndiModifications jm) throws CommunicationExceptio
 			
 			return false;
 		}
+		finally {
+			// clear requestControls for future use of the JNDI context
+			try {
+				ctx.setRequestControls(null);
```
rouazana, line 935: Do you think this configuration of the LdapContext is thread-safe? It seems ctx is shared by all threads using JndiServices, no?

remk:
Do you think you can add a test for this ?

src/main/java/org/lsc/jndi/JndiServices.java
soisik: you're right, I changed the way we use this control to derive the context for updates so the initial/shared context is not impacted by the control.

soisik:
@remk : Unit test added, but since community version of OpenDJ doesn't support this control (the EE does), it's a bit limited; it just checks that the control is added or not depending on the context called.
Since it was impacted, I wrote an extra test for recursive delete as it didn't exists.

coudot:
@remk @rouazana @soisik if all is ok for you, can we merge this for 2.1.6?

rouazana (approved):

Issue #106, opened 2020-02-10
Update all dependencies for 2.2


coudot:
The goal is to update all dependencies for the next major release

jailbird777:
Not sure if it helps (if it doesn't, sorry for the spam), but when I was hacking 2.1.x to work with Java 11, I did some updating and this combination is working.  I ls'd the two directories (stock vs mine) and diff'd them:

-ant-optional-1.5.1.jar
-api-all-1.0.0-M22.jar
-avalon-framework-4.1.3.jar
-bcprov-jdk16-140.jar
-btm-2.1.0.jar
-commons-beanutils-1.7.0.jar
-commons-beanutils-core-1.7.0.jar
-commons-cli-1.2.jar
-commons-codec-1.3.jar
-commons-collections-3.2.1.jar
-commons-configuration-1.4.jar
-commons-digester-1.8.jar
-commons-io-1.4.jar
-commons-jxpath-1.2.jar
-commons-lang-2.5.jar
-commons-logging-1.1.jar
-commons-logging-api-1.0.4.jar
+ant-optional-1.5.3-1.jar
+api-all-1.0.3.jar
+avalon-framework-api-4.3.1.jar
+avalon-framework-impl-4.3.1.jar
+avalon-logkit-2.2.1.jar
+bcprov-jdk15on-164.jar
+btm-2.1.4.jar
+commons-beanutils-1.9.4.jar
+commons-cli-1.4.jar
+commons-codec-1.14.jar
+commons-collections-3.2.2.jar
+commons-configuration-1.10.jar
+commons-digester-2.1.jar
+commons-io-2.6.jar
+commons-jxpath-1.3.jar
+commons-lang-2.6.jar
+commons-logging-1.2.jar
+commons-logging-api-1.1.jar
-groovy-all-1.7.0.jar
-guava-11.0.2.jar
-hamcrest-core-1.1.jar
-hsqldb-1.8.0.10.jar
+groovy-all-2.4.19.jar
+guava-28.2-jre.jar
+hamcrest-core-2.2.jar
+hsqldb-2.5.0.jar
-javaee-api-5.jar
-javassist-3.16.1-GA.jar
-jcifs-1.2.19.jar
-jdom-b9.jar
-je-4.1.10.jar
-jline-0.9.94.jar
-jsr305-1.3.9.jar
+istack-commons-runtime-3.0.10.jar
+javaee-api-8.0.1.jar
+javassist-3.26.0-GA.jar
[...]

PR #204, opened 2022-09-27
Update logback, maven plugin compiler and javadoc

artlog:
- update maven plugins compiler and javadoc
  - javadoc upgrade comes with java8 javadoc errors - need doclint none to pass through - it is not possible to fix all since xsd generation creates invalid javadoc - fix warnings ( convert html in comment mainly < > & ... )
- fix some wrong throws as javadoc check MalformedURLException is never thrown, remains from very old ldap library org.ietf.ldap.LDAPUrl that was later changed by unboundid then apache.directory javadoc does check it is consistent with code
- remove obsolete class PivotTransformationServiceType.java

first update of library for issue #106


soisik (changes requested):
In src/main/java/org/lsc/beans/IBean.java, line 65, the closing </p> was replaced with an opening <p>, which seems to be a mistake.

artlog:
fixed </p> and cleaned up with a script html escaping of mails

soisik (approved):

Issue #186, opened 2022-05-06
Wrong password converted for AD


kadogo:
Hello

I notice that the password "1aDhgtryjhddç" doesn't work in AD when synchronized with LSC.

lsc.xml

```
# grep 'getUnicode' lsc.xml 
            <string>AD.getUnicodePwd("1aDhgtryjhddç")</string>
```

lsc.ldif

```
# grep unicode lsc.ldif 
replace: unicodePwd
unicodePwd:: IgAxAGEARABoAGcAdAByAHkAagBoAGQAZADvv70AIgA=
```

When I convert the password manually with bash it is working like expected

```
# printf '"%s"' "1aDhgtryjhddç" | iconv --to utf-16le | base64
IgAxAGEARABoAGcAdAByAHkAagBoAGQAZADnACIA
```

We notice clearly that the results are not the same between LSC and my bash command

Cheers

artlog:
taking a look...

artlog:
@kadogo can you provide first lines of lsc.xml , to check for <?xml  encoding = line .
If content of lsc.xml is not encoded according to encoding such issue might happen.
input char won't be recognized and will be translated as invalid '�' utf16-le : efbf bd00  by java library.

kadogo:
Hello @artlog 

I don't have any line that mention the encoding in my lsc.xml files.

```
<?xml version="1.0" ?>
<lsc xmlns="http://lsc-project.org/XSD/lsc-core-2.1.xsd" revision="0"> 
...
```

Cheers

artlog:
@kadogo 
Do you know what encoding was used for this file ? 
You could  try  changing first line to  <?xml version="1.0" encoding= "UTF-8" ?>  or   <?xml version="1.0" encoding= ISO-8859-1" ?>

kadogo:
Hello @artlog 

```
# file /etc/lsc/lsc.xml 
/etc/lsc/lsc.xml: XML 1.0 document, ASCII text
```

I tried with the 2 lines (UTF-8 and ISO-8859-1) and no succes.

Cheers

artlog:
It is because is it not unicode :

creating a dummy xml file test.xml
```
<?xml version="1.0" ?>
<lsc xmlns="http://lsc-project.org/XSD/lsc-core-2.1.xsd" revision="0"> 
  <string>AD.getUnicodePwd("1aDhgtryjhddç")</string>
</lsc>
```

```
$ file test.xml
[...]

PR #206, opened 2022-09-28
Ad unicode pwd auto encode

artlog:
Tentative to encode password transparently when using ActiveDirectoryDstService.
- To test -
- 
AD.getUnicodePwd() does nothing anymore.

### DIFF
--- a/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java
+++ b/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java
@@ -45,6 +45,7 @@
  */
 package org.lsc.jndi;
 
+import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -66,11 +67,15 @@ import org.lsc.configuration.TaskType;
 import org.lsc.exception.LscServiceConfigurationException;
 import org.lsc.exception.LscServiceException;
 import org.lsc.utils.CaseIgnoreStringHashMap;
+import org.lsc.utils.directory.AD;
 
 /**
  * A custom JNDI service to translate modifications on the user's "memberOf"
  * attribute to modifications on the "member" attribute of each groups. This is
  * the way to apply changes on groups with ActiveDirectory.
+ *
+ * Support for UnicodePwd attribute encoding from cleartext.
+ *
  * @author St&eacute;phane Bond &lt;&gt;
  */
 public class ActiveDirectoryDstService extends SimpleJndiDstService {
@@ -78,6 +83,8 @@ public class ActiveDirectoryDstService extends SimpleJndiDstService {
 	final String MEMBER_OF_ATTR = "memberOf";
 
 	final String GROUP_MEMBER_ATTR = "member";
+
+	final String UNICODE_PWD_ATTR = "UnicodePwd";
 
 	public ActiveDirectoryDstService(TaskType task)
 			throws LscServiceConfigurationException {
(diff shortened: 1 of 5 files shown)


artlog:
intend to fix #186

artlog:
I fear ActiveDirectoryDstService is not in use anymore, i yet didn't find out how to configure it , would need a specific TaskType ?

coudot:
Hello,

I tried to test these changes on a server where I can reproduce the issue.


Here is the patch I applied on 2.1 branch:
```
diff --git a/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java b/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java
index 09b6acd8..b5836d40 100644
--- a/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java
+++ b/src/main/java/org/lsc/jndi/ActiveDirectoryDstService.java
@@ -45,6 +45,7 @@
  */
 package org.lsc.jndi;
 
+import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -66,11 +67,15 @@ import org.lsc.configuration.TaskType;
 import org.lsc.exception.LscServiceConfigurationException;
 import org.lsc.exception.LscServiceException;
 import org.lsc.utils.CaseIgnoreStringHashMap;
+import org.lsc.utils.directory.AD;
 
 /**
  * A custom JNDI service to translate modifications on the user's "memberOf"
  * attribute to modifications on the "member" attribute of each groups. This is
  * the way to apply changes on groups with ActiveDirectory.
+ *
+ * Support for UnicodePwd attribute encoding from cleartext.
+ *
  * @author St&eacute;phane Bond &lt;&gt;
  */
 public class ActiveDirectoryDstService extends SimpleJndiDstService {
@@ -79,6 +84,8 @@ public class ActiveDirectoryDstService extends SimpleJndiDstService {
 
        final String GROUP_MEMBER_ATTR = "member";
 
+    final String UNICODE_PWD_ATTR = "UnicodePwd";
+
        public ActiveDirectoryDstService(TaskType task)
                        throws LscServiceConfigurationException {
                super(task);
@@ -101,10 +108,44 @@ public class ActiveDirectoryDstService extends SimpleJndiDstService {
                        }
                }
 
+               // handle "UnicodePwd" encoding, apply changes on it.
+               for (int i = 0; i < lm.getLscAttributeModifications().size(); i++) {
+                       LscDatasetModification dm = lm.getLscAttributeModifications()
+                                       .get(i);
+                       if (dm.getAttributeName().equalsIgnoreCase(UNICODE_PWD_ATTR))
+                       {
+
+                               if ( dm.getValues().size() > 0 ) {
+                                       // get only first value, why would there be multiple password values ?
+                                       String password = (String) dm.getValues().get(0);
+                                       try {
+                                               // whole point is that java String have fixed internal encoding
+                                               // so ldap attribute encoding should be done lately from byte[]
+                                               byte[] pwdArray = AD.getUnicodePwdEncoded(password);
+                                               List<ModificationItem> modificationItems = new ArrayList<ModificationItem>();
+                                               modificationItems.add(new ModificationItem(DirContext.REPLACE_ATTRIBUTE, new BasicAttribute(UNICODE_PWD_ATTR, pwdArray)));
+                                               JndiModifications jndiModifications = new JndiModifications(JndiModificationType.getFromLscModificationType(lm.getOperation()), lm.getTaskName());
+                                               jndiModifications.setDistinguishName(lm.getMainIdentifier());
+                                               jndiModifications.setModificationItems(modificationItems);
+                                               success &= jndiServices.apply(jndiModifications);
+                                       }
+                                       catch(UnsupportedEncodingException unsupportedEncoding)
+                                       {
+                                               throw new LscServiceException(unsupportedEncoding);
+                                       }
+                                       catch (CommunicationException e) {
+                                               throw new LscServiceException(e);
+                                       }
+                                       lm.getLscAttributeModifications().remove(i);
+                                       break;
+                               }
+                       }
+               }
+
                // Apply regular changes
                if (lm.getLscAttributeModifications().size() > 0
                                || lm.getNewMainIdentifier() != null) {
-                       success = super.apply(lm);
+                       success &= super.apply(lm);
                }
 
                // Apply changes on memberships
diff --git a/src/main/java/org/lsc/utils/directory/AD.java b/src/main/java/org/lsc/utils/directory/AD.java
index b6e01a3d..aeb38e9c 100644
--- a/src/main/java/org/lsc/utils/directory/AD.java
+++ b/src/main/java/org/lsc/utils/directory/AD.java
@@ -115,6 +115,20 @@ public class AD {
                }
        }
 
+       /**
+        * Kept for legacy reason, don't use in any new development
+        *
+        * @deprecated UnicodePwd attribute is automatically encoded within ActiveDirectoryDstService
+        *             this method does nothing
+        *             to find actual encoding see #getUnicodePwdEncoded(String password)
+        */
+       public static String getUnicodePwd(String password)  throws UnsupportedEncodingException {
+
+               // don't do anything
+               return password;
+       }
+
+
        /**
         * Encode a password so that it can be updated in Active Directory
         * in the field unicodePwd.
@@ -123,9 +137,9 @@ public class AD {
         * @return The value to write in AD's unicodePwd attribute
         * @throws UnsupportedEncodingException
         */
-       public static String getUnicodePwd(String password) throws UnsupportedEncodingException {
+       public static byte[] getUnicodePwdEncoded(String password) throws UnsupportedEncodingException {
                String quotedPassword = "\"" + password + "\"";
-               return new String(quotedPassword.getBytes("UTF-16LE"));
+               return quotedPassword.getBytes("UTF-16LE");
        }
```


But I'm not able to see where in lsc.xml I can tell LSC to use org.lsc.jndi.ActiveDirectoryDstService

@davidcoutadeur @soisik @rouazana any idea?

artlog:
I don't know how  to use ActiveDirectoryDstService.
new method AD.getUnicodePwdEncoded(String password) should return a byte[] that is what we would like since problem of AD.getUnicodePwd(String password) is the encoding of String and can't be fixed as a String.
to unicodePWd is in fact a binary field since it does not use standard String encoding.

coudot:
Fixed by #267

Issue #133, opened 2020-09-25
Java 11+ javascript engine compatibility


rouazana:
Java 11 deprecated Nashorn, after the removal of Rhino. As there will be no more Javascript Engine in the JDK, we should migrate to an external one, for example GraalVM Javascript.
It will also allow us to support new JS capabilities from the ECMAScript 2019 norm.

artlog:
let's try graalVM https://www.graalvm.org/ ... working on.

rouazana:
Hello!

Thanks for looking at this.

For your information Nashorn is finally also available as an OpenJDK project : https://github.com/openjdk/nashorn

coudot:
Two things:
* Test with recent java versions (11, 17, ...) that default js code is well interpreted
* Add in documentation for 2.2 that graalvm.js can be used

coudot:
@abpai94 could you try to create a unit test for graalvm?

abpai94:
Documentation for `graal.js` compatible JVMs and their versions:
* `openjdk:v11`
* `azul:v13`
* `corretto:v15`
* `corretto:v16`
* `corretto:v18`
* `corretto:v19`
* `corretto:v20`
* `graalvm-ce:v21`
* `graalvm-jdk:v21`
* `openjdk:v22`
* `graalvm-jdk:v22`
* `graalvm-ce:v22`

Incompatible:
* `openjdk:v1.8`
* `graalvm-ce:v17`
* `graalvm-jdk:v17`

davidcoutadeur:
This issue has been resolved in #340 (and in #348 for documentation)

I think there is nothing more to do in this issue, so I close it.

PR #208, opened 2022-10-03
javascript graaljs support and rhino update

artlog:
- from java 11 add graaljs supported by graalvm.
- graalvm 22.2.0
- default implementation when no scripting language is provided
- graalsjs has full access to java code and work in nashorn compatibility
- use 'gj' to explictly require graaljs ( and not gr that is groovy ).
- bump rhino to 1.7.14.


artlog:
to address  #133 graaljs support

soisik (approved):

Issue #67, opened 2018-07-17
Order of task execution


dean-radish:
When I have multiple tasks in a job then they execute in alphabetical order of the task name, not in the order defined in the XML. Is this expected behavior?
This happens if I use the  **all** keyword or a list of task names

dean-radish:
See also https://lists.lsc-project.org/pipermail/lsc-users/2015-February/002492.html

dean-radish:
I'm happy to create a pull request with a fix if you agree that they should execute in the order defined in the XML file

coudot:
Why not, but you can also choose the order with lsc -s task1,task2,task3 ...

See https://lsc-project.org/documentation/latest/execution/start

coudot:
We won't modify the default behavior but I will update documentation to better explain how to fix task order

coudot:
Documentation updated

davidcoutadeur:
Seems the task order is not enforced at all.

For example:

bin/lsc -f etc/ -sMySyncTask3,MySyncTask2,MySyncTask1 -t 1
```
juin 19 16:35:03 - INFO  - Starting sync for MySyncTask1
juin 19 16:35:04 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 19 16:35:04 - INFO  - Starting sync for MySyncTask2
juin 19 16:35:05 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 19 16:35:05 - INFO  - Starting sync for MySyncTask3
juin 19 16:35:05 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
```

Seems the task order is determined by alpha-numeric sort of the task in lsc.xml

coudot:
The order is now correctly set with the command line parameters. The order will never be the one of the lsc.xml file.

PR #218, opened 2023-06-19
enforce order of user-defined tasks at launch

davidcoutadeur:
See #67, and especially https://github.com/lsc-project/lsc/issues/67#issuecomment-1597310928

### DIFF
--- a/src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
+++ b/src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
@@ -1,0 +1,190 @@
+/*
+ ****************************************************************************
+ * Ldap Synchronization Connector provides tools to synchronize
+ * electronic identities from a list of data sources including
+ * any database with a JDBC connector, another LDAP directory,
+ * flat files...
+ *
+ *                  ==LICENSE NOTICE==
+ * 
+ * Copyright (c) 2008 - 2011 LSC Project 
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the LSC Project nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+ * IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER
+ * OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *                  ==LICENSE NOTICE==
+ *
+ *               (c) 2008 - 2011 LSC Project
+ *         Sebastien Bahloul &lt;seb@lsc-project.org&gt;
+ *         Thomas Chemineau &lt;thomas@lsc-project.org&gt;
+ *         Jonathan Clarke &lt;jon@lsc-project.org&gt;
+ *         Remy-Christophe Schermesser &lt;rcs@lsc-project.org&gt;
+ ****************************************************************************
+ */
+package org.lsc;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import javax.naming.NamingException;
+import javax.naming.directory.Attribute;
+import javax.naming.directory.SearchResult;
+
+import org.apache.commons.codec.binary.Base64;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.lsc.beans.IBean;
+import org.lsc.configuration.LscConfiguration;
+import org.lsc.exception.LscServiceException;
+import org.lsc.jndi.SimpleJndiSrcService;
+import org.lsc.service.IService;
+import org.lsc.utils.directory.LDAP;
+
+/**
+ * This test case attempts to reproduce a typical ldap2ldap setup via
+ * SimpleSynchronize. This relies on settings in lsc.properties from the test
+ * resources, and entries in the local OpenDS directory. testSync() performs a
+ * synchronization between two branches of the local LDAP server, and should
+ * perform 3 operations : 1 ADD, 1 MODRDN, 1 MODIFY. The
+ * 
+ * @author Jonathan Clarke &ltjonathan@phillipoux.net&gt;
+ */
+public class Ldap2LdapOrderedSyncTest extends CommonLdapSyncTest {
+
+
+	@Before
+	public void setup() {
+		LscConfiguration.reset();
+		LscConfiguration.getInstance();
+		Assert.assertNotNull(LscConfiguration.getConnection("src-ldap"));
+		Assert.assertNotNull(LscConfiguration.getConnection("dst-ldap"));
+		reloadJndiConnections();
+	}
+
+	@Test
+	public final void testLdap2LdapOrderedSyncTest() throws Exception {
+
+		// make sure the contents of the directory are as we expect to begin with
+		List<String> user_tasks = Arrays.asList("ldap2ldapOrderedTestTask3", "ldap2ldapOrderedTestTask2", "ldap2ldapOrderedTestTask1");
+
+		// perform the sync
+		launchSyncCleanTask(user_tasks, false, true, false);
+
+		// check the results of the synchronization
+		reloadJndiConnections();
+		checkSyncResults();
+	}
+
+	private final void checkSyncResults() throws Exception {
+		List<String> attributeValues = null;
+
+
+		// check Operation order
+
+	}
+
+	public static void launchSyncCleanTask(List<String> tasks, boolean doAsync, boolean doSync,
+					boolean doClean) throws Exception {
+		// initialize required stuff
+		SimpleSynchronize sync = new SimpleSynchronize();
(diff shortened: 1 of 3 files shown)


rouazana (commented):
Hello,

Thanks for trying to fix issue! I've just experienced it by the way.

The easiest workaround is to match the order of your task with the alphabetic order:
`1-first-task, 2-second-task, 3-third-task, etc.`

But I agree we should change the code for this so that in the next version tasks are executed in the same order they are submitted by the user.

I guess your fix should work but it seems a little over complicated. I think a more simpler solution would be [at this line](https://github.com/lsc-project/lsc/commit/515e459549b8f96b3dac47a42a7af8ccc2a286bc#diff-d5b6d928746bbfe34818c3bd0b85a5a23137c0d41e3a8ee98296422091e9bcf8R97) to replace TreeMap by a LinkedHashMap. It should do the job in a much simpler way.

Could you check it's working as expected?

I think it would also be worth to add a test to check and enforce this behavior in the future.

Thank you,
Raphaël.

davidcoutadeur:
Hi!
Thanks for your help. :) I'll try your suggestion quite soon.

davidcoutadeur:
For info, I have also tested your suggestion:

```
diff --git a/src/main/java/org/lsc/SimpleSynchronize.java b/src/main/java/org/lsc/SimpleSynchronize.java
index aa8f6d63..6943471d 100644
--- a/src/main/java/org/lsc/SimpleSynchronize.java
+++ b/src/main/java/org/lsc/SimpleSynchronize.java
@@ -57,6 +57,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.LinkedHashMap;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.lsc.beans.IBean;
@@ -95,7 +96,7 @@ public class SimpleSynchronize extends AbstractSynchronize {
        public SimpleSynchronize() {
                super();
                setThreads(5); 
-               cache = new TreeMap<String, Task>();
+               cache = new LinkedHashMap<String, Task>();
        }
        
        public void init() throws LscConfigurationException {

```

For me, it does not work, because when I launch:

```
lsc -f ~/Documents/lsc/etc -sMySyncTask3,MySyncTask2,MySyncTask1 -cMySyncTask3,MySyncTask2,MySyncTask1 -t 1
```

The order is not enforced:

```
juin 29 11:03:44 - INFO  - Starting sync for MySyncTask1
juin 29 11:03:45 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 29 11:03:45 - INFO  - Starting clean for MySyncTask1
juin 29 11:03:45 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 29 11:03:45 - INFO  - Starting sync for MySyncTask2
juin 29 11:03:45 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 29 11:03:45 - INFO  - Starting clean for MySyncTask2
juin 29 11:03:45 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 29 11:03:45 - INFO  - Starting sync for MySyncTask3
juin 29 11:03:46 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
juin 29 11:03:46 - INFO  - Starting clean for MySyncTask3
juin 29 11:03:46 - INFO  - All entries: 8, to modify entries: 0, successfully modified entries: 0, errors: 0
```

But maybe I didn't understand your point?

rouazana:
Your are probably right. I guess my proposed fix takes the tasks in the order they are in the configuration file, but still not as specified on the command line.

rouazana:
I you can add a test it would be great, else I'm fine with your fix.

Also I would have argued that it would have been better to extract methods instead of copying code, but it seems it will be hard given all the boolean that are manipulated.

davidcoutadeur:
> Also I would have argued that it would have been better to extract methods instead of copying code, but it seems it will be hard given all the boolean that are manipulated.

I really tried believe me :)

I'll take care of the test

rouazana:
Just to notice: your fix introduces a slightly different behavior.
For example with `-s t1,t2 -c t1,t2` before LSC was doing:
```
sync t1, clean t1, sync t2, clean t2
```
If I got it right with your fix it will do:
```
sync t1, sync t2, clean t1, clean t2
```

I'm not sure if an issue, but in case you did not notice it...

rouazana:
Also here should be a slightly simpler fix:
```
diff --git a/src/main/java/org/lsc/SimpleSynchronize.java b/src/main/java/org/lsc/SimpleSynchronize.java
index 4e5a0835..61184599 100644
--- a/src/main/java/org/lsc/SimpleSynchronize.java
+++ b/src/main/java/org/lsc/SimpleSynchronize.java
@@ -57,6 +57,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.lsc.beans.IBean;
@@ -138,16 +139,10 @@ public class SimpleSynchronize extends AbstractSynchronize {
 	 */
 	public final boolean launch(final List<String> asyncTasks, final List<String> syncTasks,
 					final List<String> cleanTasks) throws Exception {
-		Boolean foundATask = false;
+		boolean foundATask = false;
 		boolean canClose = true;
 		boolean launchResult = true;
 
-		// Get the list of defined tasks from LSC properties
-		// Iterate on each task
-		boolean isASyncTaskAll = asyncTasks.contains(ALL_TASKS_KEYWORD);
-		boolean isSyncTaskAll = syncTasks.contains(ALL_TASKS_KEYWORD);
-		boolean isCleanTaskAll = cleanTasks.contains(ALL_TASKS_KEYWORD);
-		
 		if(getTasksName() == null) {
 			return false;
 		} else if(getTasks().length == 0) {
@@ -156,43 +151,43 @@ public class SimpleSynchronize extends AbstractSynchronize {
 		if(!asyncTasks.isEmpty()) {
             LscServerImpl.startJmx(this);
 		}
-		
-		for (Task task: cache.values()) {
 
-			// Launch the task either if explicitly specified or if "all" magic keyword used
-			if (isSyncTaskAll || syncTasks.contains(task.getName())) {
-				foundATask = true;
+		Collection<Task> asyncTasksToLaunch = userOrderedTask(asyncTasks, asyncTasks.contains(ALL_TASKS_KEYWORD));
+		Collection<Task> syncTasksToLaunch = userOrderedTask(syncTasks, syncTasks.contains(ALL_TASKS_KEYWORD));
+		Collection<Task> cleanTasksToLaunch = userOrderedTask(cleanTasks, cleanTasks.contains(ALL_TASKS_KEYWORD));
 
-				if (!launchTask(task, Task.Mode.sync)) {
-					launchResult = false;
-				} else {
-					if(task.getSyncHook() != null && task.getSyncHook() != "") {
-						runPostHook(task.getName(), task.getSyncHook(), task.getTaskType());
-					}
+		for (Task task: syncTasksToLaunch) {
+			foundATask = true;
+
+			if (!launchTask(task, Task.Mode.sync)) {
+				launchResult = false;
+			} else {
+				if(task.getSyncHook() != null && task.getSyncHook() != "") {
+					runPostHook(task.getName(), task.getSyncHook(), task.getTaskType());
 				}
 			}
-			if (isCleanTaskAll || cleanTasks.contains(task.getName())) {
-				foundATask = true;
+		}
+		for (Task task: cleanTasksToLaunch) {
+			foundATask = true;
 
-				if (!launchTask(task, Task.Mode.clean)) {
-					launchResult = false;
-				} else {
-					if(task.getCleanHook() != null && task.getCleanHook() != "") {
-						runPostHook(task.getName(), task.getCleanHook(), task.getTaskType());
-					}
+			if (!launchTask(task, Task.Mode.clean)) {
+				launchResult = false;
+			} else {
+				if(task.getCleanHook() != null && task.getCleanHook() != "") {
+					runPostHook(task.getName(), task.getCleanHook(), task.getTaskType());
 				}
 			}
-			if (isASyncTaskAll || asyncTasks.contains(task.getName())) {
-				foundATask = true;
-				
-				canClose = false;
+		}
+		for (Task task: asyncTasksToLaunch) {
+			foundATask = true;
 
-				if(!launchTask(task, Task.Mode.async)) {
-					launchResult = false;
-				}
+			canClose = false;
+
+			if(!launchTask(task, Task.Mode.async)) {
+				launchResult = false;
 			}
 		}
-		
+
 		if (canClose) {
 			close();
 		}
@@ -205,6 +200,18 @@ public class SimpleSynchronize extends AbstractSynchronize {
 		return launchResult;
 	}
 
+	private Collection<Task> userOrderedTask(List<String> userValues, boolean all) {
+		Collection<Task> allTasks = cache.values();
+		if (all) {
+			return allTasks;
+		}
+		Collection<String> allTaskNames = allTasks.stream().map(Task::getName).collect(Collectors.toList());
+		return userValues.stream()
+				.filter(allTaskNames::contains)
+				.map(cache::get)
+				.collect(Collectors.toList());
+	}
+
 	/**
 	 * Launch a task. Call this for once each task type and task mode.
 	 *
```
(exactly the same behavior than yours)

davidcoutadeur:
> Just to notice: your fix introduces a slightly different behavior. For example with `-s t1,t2 -c t1,t2` before LSC was doing:
> 
> ```
> sync t1, clean t1, sync t2, clean t2
> ```
> 
> If I got it right with your fix it will do:
> 
> ```
> sync t1, sync t2, clean t1, clean t2
> ```
> 
> I'm not sure if an issue, but in case you did not notice it...

Yes, I already had that in mind. It does not seem a big change, and seems quite logical.

Anyway, as this is a modification of behaviour, there is a discussion about "in which version we could apply this fix". I mean that really enforcing the task order *is* a real modification of the behaviour, even if the doc says the contrary.

Thus it could be applied to the future major release maybe?

davidcoutadeur:
> ...
> (exactly the same behavior than yours)

Seems more elegant indeed. I'll try this and add the test this week.

davidcoutadeur:
@rouazana I have applied your suggestion + written the test for enforcing the task order.

Could you review the test? If it is ok for you, I think we can merge.

src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
```diff
@@ -0,0 +1,190 @@
...
+		// check the results of the synchronization
+		reloadJndiConnections();
+		checkSyncResults();
+	}
+
+	private final void checkSyncResults() throws Exception {
+		List<String> attributeValues = null;
+
+
+		// check Operation order
```
rouazana, line 116: it seems your test is not checking anything ^^

src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
```diff
@@ -0,0 +1,190 @@
...
+import org.lsc.jndi.SimpleJndiSrcService;
+import org.lsc.service.IService;
+import org.lsc.utils.directory.LDAP;
+
+/**
+ * This test case attempts to reproduce a typical ldap2ldap setup via
+ * SimpleSynchronize. This relies on settings in lsc.properties from the test
+ * resources, and entries in the local OpenDS directory. testSync() performs a
+ * synchronization between two branches of the local LDAP server, and should
+ * perform 3 operations : 1 ADD, 1 MODRDN, 1 MODIFY. The
```
rouazana, line 82: copy/pasted comment?

rouazana (commented):
it seems your test is not ensuring the order is enforced.

src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
davidcoutadeur: The test itself just launch the tasks in a specific order (3,2,1), while these tasks are defined in an other order in lsc.xml (1,2,3).
But when the tasks are executed in the order (1,2,3), it ends inevitably in an error. (see lsc.xml content)
The only way the tasks can succeed is to execute in order (3,2,1)

src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
davidcoutadeur: Yes, duplicated from `Ldap2LdapSyncTest.java`

src/test/java/org/lsc/Ldap2LdapOrderedSyncTest.java
davidcoutadeur: The other idea I had was to capture the lsc output to check the real order execution, but it seemed more complex to me.

rouazana: ohhhh I totally missed this.

I think it's ok if you add two comments:
1. replace the [main comment](https://github.com/lsc-project/lsc/pull/218#discussion_r1254837017) by an explanation
2. replace `// check Operation order` by a little explanation telling why it fails

davidcoutadeur: I have fixed your 2 last points. I will merge the PR

rouazana:
Great thank you!

PR #226, opened 2023-09-13
Fix inaccurate log message when loading script engine

amottier:
A script engine (e.g. "Oracle Nashorn") is incorrectly reported as unsupported in debug log message.

If we take as an example the "Oracle Nashorn" script engine, it has the following short names: nashorn, Nashorn, js, JS, JavaScript, javascript, ECMAScript, ecmascript.
With the current implementation, the script engine will be, by default, considered to be supported and if, while iterating on the list of short names, a short name is *not one of*: js, groovy, grall.js, the engine will be considered to be *not supported*. So with the example of "Oracle Nashorn" it will be reported as not supported event if one of its short name is `js` and it is actually loaded (the issue is only related to the log message).

This commit revert the logic: a script engine is by default unsupported and if one of its short names match one of js, groovy, grall.js it will be considered supported.

### DIFF
--- a/src/main/java/org/lsc/utils/ScriptingEvaluator.java
+++ b/src/main/java/org/lsc/utils/ScriptingEvaluator.java
@@ -43,15 +43,17 @@ public class ScriptingEvaluator {
 		instancesTypeCache = new HashMap<String, ScriptableEvaluator>();
 		List<ScriptEngineFactory> factories = mgr.getEngineFactories();
 		for (ScriptEngineFactory sef : factories) {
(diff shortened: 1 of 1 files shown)


rouazana (approved):
Looks fine at first sight, thank you!

amottier:
@rouazana thanks for the review. Should I do anything else to allow the merge of this pull request?

rouazana:
Sorry for the delay, it's merged now :)

amottier:
Thansk a lot @rouazana :+1:

PR #228, opened 2023-09-13
Specify encoding to use for .properties files

amottier:
`maven-resources-plugin` is using `project.build.sourceEncoding` to specify encoding to use when copying filtered resources (see https://maven.apache.org/plugins/maven-resources-plugin/examples/encoding.html). `.properties` files are encoded in `ISO-8859-1` (see [Java 8 API documentation](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html), note that by default, since Java 9, Java runtime try to [read properties file as `UTF-8` encoded](https://docs.oracle.com/en/java/javase/11/intl/internationalization-enhancements1.html#GUID-974CF488-23E8-4963-A322-82006A7A14C7) en fallback to `ISO-8859-1` if needed). To prevent issue, `maven-resources-plugin` log a warning if encoding for `.properties` files is not specified in plugin configuration (see https://maven.apache.org/plugins/maven-resources-plugin/examples/filtering-properties-files.html).

This commit set the encoding to be used for `.properties` files to `ISO-8859-1`.

The purpose of this commit is solely to avoid a `maven-resources-plugin` as the project doesn't currently include `.properties` files.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
 				<artifactId>maven-resources-plugin</artifactId>
 				<version>3.3.1</version>
 				<configuration>
-					<propertiesEncoding>ISO-8859-1</propertiesEncoding><!-- To avoid warning during plugin execution. The project currently doesn't include properties files. -->
+					<propertiesEncoding>UTF-8</propertiesEncoding><!-- To avoid warning during plugin execution. -->
 				</configuration>
 			</plugin>
 
--- a/pom.xml
+++ b/pom.xml
@@ -170,6 +170,14 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-resources-plugin</artifactId>
+				<version>3.3.1</version>
+				<configuration>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -170,6 +170,14 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-resources-plugin</artifactId>
+				<version>3.3.1</version>
+				<configuration>
+					<propertiesEncoding>ISO-8859-1</propertiesEncoding><!-- To avoid warning during plugin execution. The project currently doesn't include properties files. -->
```
rouazana, line 178: Wouldn't it be better to set it to UTF-8?

By the way do you know why it's not already the default?

amottier: As the project currently target Java 8 I choose to use `ISO-8859-1` over `UTF-8`. But as the  LSC doesn't include any properties file the configuration here does not actually matters and it safe to set the value to `UTF-8`.

amottier: I totally agree with you and change the encoding from `ISO-8859-1` to `UTF-8` in a new commit.

I don't know how but I didn't find the properties files during my first scan of the project files, sorry.

rouazana: No problem, than you very much for the change!

Issue #50, opened 2018-04-20
Make a systemd script


davidcoutadeur:
I don't think this issue was proposed previously, so I post it here.
It was first proposed by francesco.malvezzi@unimore.it in the mailing list (2018-02-13 13:50 GMT+01:00)

```
[Unit]
Description=LSC sync broker
After=network.target

[Service]
PrivateTmp=true
PIDFile=/run/lsc.pid
Type=oneshot
#Environment=JAVA_HOME=/usr/jvm/lib/latest
User=lsc
Group=lsc
ExecStart=/opt/lsc/bin/lsc -a all -f /opt/lsc/etc

[Install]
WantedBy=multi-user.target
```

Need to be tested.

coudot:
Check that we can read logs with journalctl

bendem:
`PIDFile` is not required with `Type=oneshot`. We run lsc from systemd with a few more hardening options (I believe `ExecStart` and `ReadWritePaths` are specific to our setup, they should be made generic).

```ini
[root@ldap-test-01 lsc]# systemctl cat lsc-main
# /etc/systemd/system/lsc-main.service
#
# Ansible managed
#
[Unit]
Description=Run LSC jobs | main

[Service]
Type=oneshot
ExecStart=/etc/lsc/run-main
User=lsc
Group=lsc

PrivateTmp=true
PrivateDevices=true
ProtectSystem=strict
ProtectKernelTunables=yes
ProtectControlGroups=yes
ReadWritePaths=/var/log/lsc
UMask=0007

[Install]
WantedBy=multi-user.target
```

artlog:
taking a look.

davidcoutadeur:
I have proposed a systemd example in https://github.com/lsc-project/lsc/pull/230
- it runs `lsc -f /etc/lsc -s all`
- stdout logs are displayed in `journalctl -xeu lsc.service`

Certainly not perfect because the admin still need to adapt the synchronization options (-s, -c, -a)

coudot:
Solution to test: use the standard way to override a systemd service : https://askubuntu.com/questions/659267/how-do-i-override-or-configure-systemd-services

davidcoutadeur:
I have added a new commit (6b9afd78) for:
- adding debian package recipes
- adding RHEL package recipes
- fixing some minor things
- renaming the systemd service names and corresponding default files, which I didn't find meaningful:

lsc-oneshot -> lsc-sync
lsc -> lsc-async
[...]

PR #230, opened 2023-10-06
Make a systemd script (#50)

davidcoutadeur:


davidcoutadeur:
This MR is obsolete and replace by #260

Issue #43, opened 2018-02-27
objectGUID as pivot attribute?


jeremypng:
I saw there was an old request to make the objectGUID convert to a string so that it could be used as the pivot attribute. Did that ever get handled? I see in the docs the it says the pivot has to be a string... I'm trying to setup a synchronization that maintains user connections across email address and UPN changes.

coudot:
As far as I know this is still not possible. But I let @rouazana or @soisik confirm.

rouazana:
If I remember correctly, you should be able to use it as a binary pivot, without conversion:
https://github.com/lsc-project/lsc/blob/master/src/test/java/org/lsc/Ldap2LdapBinaryPivotSyncTest.java
If you need conversion, you should have a look at trunk which implement pivot conversion:
https://github.com/lsc-project/lsc/blob/master/src/test/java/org/lsc/Ldap2LdapBinaryPivotWithStringConversionSyncTest.java

dswitzer:
I'm running into the same issue. I've looked at the source code links @rouazana link to. If I'm reading things correctly, I really need to go look at https://github.com/lsc-project/lsc/blob/master/src/test/resources/etc/lsc.xml to see how the configuration was implemented.

If I look at the Ldap2LdapBinaryPivotSyncTest.java example, it doesn't show needing to use the `<binaryAttributes>` element to define the attribute. 

However in our testing using the `objectGUID` doesn't work as a pivot table with or without the `<binaryAttributes>` element to define the attribute. The `<binaryAttributes>` gets us past a few errors, but it doesn't seem to get encoded correctly when passed to the filter queries. 

Should this be working using 2.1.4 or was this something just fixed in master?

Also, is there somewhere the better explains what the "conversion" example?

I'm guessing it's the `<pivotTransformation>` block in the 'ldap2ldapBinary2StringTestTask' test configuration that's doing the conversion, but the way I read the XML is it's transforming the pivot attribute from a field called "personalSignature" to one called "favouriteDrink", but for our project we're syncing to a database, not another LDAP server. The transform code also looks suspect to me. Doesn't really seem to be doing anything other than reformatting the each number into a delimited sequence (which may is the correct thing).

coudot:
@soisik please provide utility code to convert objectGUID value

coudot:
Some code example here: https://stackoverflow.com/a/18384620

artlog:
taking a look...

coudot:
Code found on the mailing list:
```java
[...]

PR #233, opened 2023-10-16
Added utility method + unit test to convert a byte[] objectGUID attri…

soisik:
…bute from AD to a UUID string representation #43


soisik:
Closing this to open another one shortly, coming from a specific branch and not worteks:master.

Issue #108, opened 2020-02-10
--threads parameter used only in sync phase


rouazana:
In clean phase the --threads parameter is ignored.

It could be useful in some cases, for example when the source service has a big latence. Getting each element on the source service could be really faster if it could be parallelized.

coudot:
@soisik @xavierba what is the status of this issue, is it fixed?

PR #235, opened 2023-10-18
Use multithreading for clean phase using --thread parameter (#108)

soisik:
- Use multithreading for clean phase using --thread parameter (#108)
- Refactored AbstractSynchronize (removed inner classes)
- Fixed ident


rouazana (approved):
It's really hard to review as you mixed refactoring, extracting and introducing the thread pool for the clean phase.
But at first sight it seems good!

Issue #237, opened 2023-10-18
Automatic build is broken


rouazana:
Hello,

The Travis build is not available anymore. It's grey on [the main page](https://github.com/lsc-project/lsc) and there is no automatic build on the last pull requests. It makes me dubious merging the last pull requests.

coudot:
Yes Travis can't be used anymore for free.

We need to set up github actions

rouazana:
Ok thanks, I thought it was still available for Open Source: https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source

coudot:
Yes but you need to ask for build credits and renew the operation everytime you need more.

See #238

PR #238, opened 2023-10-18
Add github action

coudot:
See #237

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -29,11 +29,3 @@ jobs:
         cache: maven
     - name: Build with Maven
       run: mvn -B -Popendj package --file pom.xml
-    - name: Set up JDK 11
-      uses: actions/setup-java@v3
-      with:
-        java-version: '11'
-        distribution: 'temurin'
-        cache: maven
-    - name: Build with Maven
-      run: mvn -B -Popendj package --file pom.xml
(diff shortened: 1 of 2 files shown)


coudot:
I added another steps with JDK 11 but test with JDK 8 are failing with:
```
Warning:       [echo] Starting OpenDJ embedded server from /home/runner/work/lsc/lsc/target/test-classes
[INFO]      [java] Unrecognized option: --add-exports=java.base/sun.security.tools.keytool=ALL-UNNAMED
[INFO]      [java] Error: Could not create the Java Virtual Machine.
[INFO]      [java] Error: A fatal exception has occurred. Program will exit.
Error:       [java] Java Result: 1
Error: The operation was canceled.
```

Not sure it is linked to my modification. What do you think?

coudot:
Same error after reverting the latest commit, should be linked to https://github.com/lsc-project/lsc/commit/2d47fa38fa09a5c7feb44fef1eae9372963a1613

Trying to run JDK 11 test before JDK 8 test to confirm this.

coudot:
Seems we could use if/else in ant: https://ant.apache.org/manual/ifunless.html

@rouazana is this something you can look at?

rouazana:
Strange, I tested the build with Java 8 without any issue.

I will have a look.

rouazana:
It should be fixed, I will rebase your PR to check.

Issue #95, opened 2019-10-02
Insecure temp file creation


linnea-s:
The sample configuration and scripts write many files to /tmp, as can be seen with fgrep -r /etc lsc. In particular bin/lsc and and etc/logback.xml, which are probably used unchanged by many users following the instructions on the website. These temporary files have known, fixed names.  This is insecure and can lead to information disclosure or overwriting of files by another user on the system. See e.g. https://cwe.mitre.org/data/definitions/377.html

I suggest defaulting to writing logs to LSC_HOME, and using a secure method for creating temporary files and directories.

coudot:
The log directory is already changed in packages (RPM/Debian).

It could indeed be changed in the source code.

coudot:
This requires changes in bin/lsc and logback.xml

coudot:
Packaging files must also be updated

davidcoutadeur:
In debian packages, I have not found in reference to tmp directory in `/usr/bin/lsc`, nor `/etc/lsc/*`
Maybe this problem is only present in source files.

coudot:
Proposed solution: create a randon tmp dir (`mktemp` with lsc prefix) for LOG_DIR. But LOG_DIR will still be overridden in RPM and DEB.

davidcoutadeur:
Not so trivial.

LOG_DIR in bin/lsc is used only for logging start and end of lsc run.

The main problem is to fix logback.xml:

```
<property name="logdir" value="/tmp/lsc/log" />
```

I am working on this.

davidcoutadeur:
Done in 77ff84bf

PR #241, opened 2023-10-19
use a temporary log directory unless configured statically (#95)

davidcoutadeur:


davidcoutadeur:
Seems all fine for me, but I could run the test because of https://github.com/lsc-project/lsc/pull/238#issuecomment-1770175306

We need to run the test when fixed.

davidcoutadeur:
Tests are successful. Everything seems fine to me!

Issue #193, opened 2022-08-22
Hook / Trigger system for add/delete/modify/modrdn


coudot:
A useful feature would be to be able to call a script (like it is done in executable plugin) but when an entry is added/modified/deleted/...

A common use case is to send a mail to user when the entry is created.

davidcoutadeur:
Seems all good to me.

For now:
- only a posthook is added
- call the script with 3 arguments: entry identifier (DN), operation (create, update, delete, changeid), operations
- operations is a json structure representing the operations on the entry

TODO: if/when merged, it is required to update the documentation to describe the behavior. (on the master branch of the documentation corresponding to 2.2)

davidcoutadeur:
The test is missing, I am working on it.

davidcoutadeur:
The MR has been merged.

The documentation is updated here: https://github.com/lsc-project/documentation/commit/613e3d9104f6e6b794998cb019c0169c4fb2c94a

All seems ok.

davidcoutadeur:
Small test regression. I don't know how I could not have noticed before. Should be solved by 62eda9f5

PR #242, opened 2023-10-23
add a create/update/changeId/delete posthook (#193)

davidcoutadeur:


soisik:
@davidcoutadeur please rebase your PR since we have merged some important refactoring commits since you have sent this.

davidcoutadeur:
> @davidcoutadeur please rebase your PR since we have merged some important refactoring commits since you have sent this.

Yes, I am working on this! Coming soon.

davidcoutadeur:
Done

davidcoutadeur:
The test is missing, I am working on it.

davidcoutadeur:
I have just added the test. Everything seems fine.

rouazana:
I have some little remarks about the code itself, but nothing very important.

Before that I would like to understand the purpose. I understand that you want to be able to launch a process after each modification/creation/deletion.
My questions are:
- Why are you using json as an argument to pass the modifications? Given how LSC works, and for consistency, I would more have expected to use LDIF, no?
- Also it seems strange to have a command argument which can be pretty big (think of changing the members of a group for example), so why not using standard input for that instead?

davidcoutadeur:
Hello

> I have some little remarks about the code itself, but nothing very important.
> 
> Before that I would like to understand the purpose. I understand that you want to be able to launch a process after each modification/creation/deletion. My questions are:
> 
>     * Why are you using json as an argument to pass the modifications? Given how LSC works, and for consistency, I would more have expected to use LDIF, no?

It is a topic open to discussion indeed.

The main reason is that I'd like to keep the generic side of the LSC object `LscModifications`. If later we have a connector to another protocol, this would permit to have a generic modification object sent in the script. In this context, JSON seemed a simple way to sent this modification object.

> 
>     * Also it seems strange to have a command argument which can be pretty big (think of changing the members of a group for example), so why not using standard input for that instead?

Good point indeed. I'll do the modification to use the standard input instead.

rouazana:
> The main reason is that I'd like to keep the generic side of the LSC object LscModifications. If later we have a connector to another protocol, this would permit to have a generic modification object sent in the script. In this context, JSON seemed a simple way to sent this modification object.

LscModifications will always be the object to pass the modifications, whatever the protocol is used in the services.
I think the LDIF format is nearer from LscModifications than JSON, hence my suggestion to use LDIF.

soisik:
> LscModifications will always be the object to pass the modifications, whatever the protocol is used in the services. I think the LDIF format is nearer from LscModifications than json, hence my suggestion to use ldif.

Sorry to step in, but we could have both, adding an option in LSC configuration to choose the input format for hooks ?
It would also be nice a feature to add to executable plugin - which use ldif as output format - to choose between ldif or json as these scripts are often calling web services, and handling data in json is easier than handling ldif for many developers.

rouazana:
Yes it could be an interesting option!

Then did you check how binary attributes are encoded in JSON?

davidcoutadeur:
I have pushed a new commit with:
- new option "outputFormat" for the LSC modifications ("ldif" or "json")
- LSC is now pushing LSC modifications as input and not argument anymore

davidcoutadeur:
> Then did you check how binary attributes are encoded in JSON?

I have made a check with jpegPhoto, which seems to be passed as base64 in LSC modification (and so in JSON)

davidcoutadeur:
Also tested with `userCertificate;binary` attribute

The data seems encoded in base64 as well:

```
[ {
  "attributeName" : "userCertificate;binary",
  "values" : [ "MIIDlzCCAn+..." ],
  "operation" : "REPLACE_VALUES"
} ]

```

src/main/java/org/lsc/Hooks.java
```diff
@@ -63,31 +64,56 @@ public class Hooks {
 	 *
 	 * return nothing
 	 */
-	public final static void postSyncHook(final String hook, final LscModifications lm) {
+	public final static void postSyncHook(final String hook, final String outputFormat, final LscModifications lm) {
```
rouazana, line 67: style: in general in Java we don't use static methods. You can construct the object in the constructor of the caller.

src/main/java/org/lsc/Hooks.java
```diff
@@ -63,31 +64,56 @@ public class Hooks {
...
 
 		if( hook != null && ! hook.equals("") )
 		{
-			// Compute json modifications
-			String jsonModifications = null;
+
+			String format = "";
+			if( outputFormat.equals("json") ) {
```
rouazana, line 73: the parsing of the outputFormat should be done previously, here you should be able to manipulate only an Enum

src/main/java/org/lsc/Hooks.java
```diff
@@ -63,31 +64,56 @@ public class Hooks {
...
 				case CREATE_OBJECT:
-					jsonModifications = getJsonModifications(lm);
-					callHook("create", hook, lm.getMainIdentifier(), jsonModifications);
+					if( format.equals("json") ) {
+						modifications = getJsonModifications(lm);
+					}
+					else {
+						modifications = LdifLayout.format(lm);
+					}
+					callHook("create", hook, lm.getMainIdentifier(), format, modifications);
```
rouazana, line 92: these lines can be easily factorized, and in this case easily integrated in the callHook method

src/main/java/org/lsc/Hooks.java
```diff
@@ -96,20 +122,32 @@ public final static void postSyncHook(final String hook, final LscModifications
 		}
 	}
 
+	/**
+	* Method calling the hook
+	*
+	* return nothing
+	*/
```
rouazana, line 129: I'm not sure such comments are really useful

src/main/java/org/lsc/beans/syncoptions/PropertiesBasedSyncOptions.java
```diff
@@ -197,6 +197,13 @@ public String getCondition(LscModificationType operation) {
 		return result;
 	}
 
+	public String getPostHookOutputFormat() {
+		if (conf.getHooks() == null || conf.getHooks().getOutputFormat() == null) {
+			return "";
```
rouazana, line 202: As said earlier, the result of this method should be an enum.
Also, I would make clearer here which is the default value.

pom.xml
```diff
@@ -872,6 +872,21 @@
 			<artifactId>xml-apis</artifactId>
 			<version>1.4.01</version>
 		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-core</artifactId>
+			<version>2.9.6</version>
```
rouazana, line 878: could you put the common version in a property?
(by the way do you really need these 3 dependencies?)

src/main/java/org/lsc/Hooks.java
```diff
@@ -0,0 +1,189 @@
...
+					identifier,
+					operationType)
+				.start();
+			}
+		}
+		catch(IOException e) {
+			StringWriter sw = new StringWriter();
+			PrintWriter pw = new PrintWriter(sw);
+			e.printStackTrace(pw);
+			LOGGER.error("Error while calling {} posthook {} with format {} for {}: {}",
```
rouazana, line 164: do you know that the LOGGER.error can directly take an exception as a parameter, and so log the stack trace without all this stuff?

src/main/java/org/lsc/Hooks.java
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	/**
+	* Method computing modifications as json
+	*
+	* @return modifications in a json String
+	*/
+	public final static String getJsonModifications(final LscModifications lm) {
+		ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
```
rouazana, line 175: nits (perf): it's recommended to declare your ObjectMapper as static for performace reasons.

src/main/java/org/lsc/Hooks.java
```diff
@@ -0,0 +1,189 @@
...
+		String json = "";
+		try {
+			json = ow.writeValueAsString(lm.getLscAttributeModifications());
+		}
+		catch(Exception e) {
+			StringWriter sw = new StringWriter();
+			PrintWriter pw = new PrintWriter(sw);
+			e.printStackTrace(pw);
+			LOGGER.error("Error while encoding LSC modifications to json", sw.toString());
```
rouazana, line 184: same remark about stack trace logging

src/main/java/org/lsc/beans/syncoptions/PropertiesBasedSyncOptions.java
```diff
@@ -196,6 +196,60 @@ public String getCondition(LscModificationType operation) {
...
+		}
+		return conf.getHooks().getOutputFormat();
+	}
+
+	public String getCreatePostHook() {
+		if (conf.getHooks() == null || conf.getHooks().getCreatePostHook() == null) {
+			return "";
```
rouazana, line 209: style: I usually prefer returning an Optional<String> to indicate there is a value or not.

But here I did not check what are the consequences, it can be a little more complicated to handle if you are not used to.

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+		// perform the sync
+		launchSyncCleanTask(sync_tasks, false, true, false);
+		launchSyncCleanTask(clean_tasks, false, false, true);
+
+		// check the results of the synchronization
+		reloadJndiConnections();
+
+		ObjectMapper mapperCreatedEntry = new ObjectMapper();
+		JsonNode expectedCreatedEntry = mapperCreatedEntry.readTree("[ { \"attributeName\" : \"objectClass\", \"values\" : [ \"inetOrgPerson\", \"person\", \"top\" ], \"operation\" : \"ADD_VALUES\" }, { \"attributeName\" : \"cn\", \"values\" : [ \"CN0001-hook\" ], \"operation\" : \"ADD_VALUES\" }, { \"attributeName\" : \"sn\", \"values\" : [ \"CN0001-hook\" ], \"operation\" : \"ADD_VALUES\" } ]");
```
rouazana, line 106: could you add a binary attribute to your test? It will allow to check for regression regarding the handling of binary attributes in hooks.

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -0,0 +1,261 @@
...
+			Scanner hookReader = new Scanner(hookFile);
+
+			while (hookReader.hasNextLine()) {
+				String data = hookReader.nextLine();
+				hookResults.add(data);
+			}
+			hookReader.close();
+			hookFile.delete(); // remove hook log
+		} catch (Exception e) {
+			fail("Error while reading hook-ldif-" + operation + ".log");
```
rouazana, line 173: you don't really need to try/catch in test, the test will fail anyway in case of unexpected exception.

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -0,0 +1,261 @@
...
+			Scanner hookReader = new Scanner(hookFile);
+
+			while (hookReader.hasNextLine()) {
+				String data = hookReader.nextLine();
+				hookResults.add(data);
+			}
+			hookReader.close();
+			hookFile.delete(); // remove hook log
+		} catch (Exception e) {
+			fail("Error while reading hook-json-" + operation + ".log");
```
rouazana, line 205: idem

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -0,0 +1,261 @@
...
+			String entry = String.join("", new ArrayList(hookResults.subList(2, hookResults.size())));
+
+			ObjectMapper mapper = new ObjectMapper();
+			JsonFactory factory = mapper.getJsonFactory();
+			JsonParser jp = factory.createJsonParser(entry);
+			try {
+				JsonNode hookOperation = mapper.readTree(jp);
+				assertEquals(hookOperation, expectedEntry);
+			}
+			catch (Exception e) {
```
rouazana, line 221: idem

rouazana (commented):
I left several style comments, but this PR is pretty good!

I like the functionality and the implementation is pretty nice and straightforward, good job!

davidcoutadeur:
test with the binary attribute is now done in d9271d4b thanks to @soisik advice

src/main/java/org/lsc/Hooks.java
```diff
@@ -58,62 +58,33 @@
...
 	/**
 	 * Method calling a postSyncHook if necessary
 	 *
 	 * return nothing
 	 */
-	public final static void postSyncHook(final String hook, final String outputFormat, final LscModifications lm) {
+	public final void postSyncHook(final Optional<String> hook, final OutputFormat outputFormat, final LscModifications lm) {
 
-		if( hook != null && ! hook.equals("") )
+		if( hook != null && hook.isPresent() )
```
rouazana, line 71: It looks better.
The principle of an optional is to not be null, so you can remove the null check.

Also if instead of a String operationType you continue using LscModificationType, the code can be very simplified by:
```
hook.ifPresent(h -> callHook(lm.getOperation(), h, lm.getMainIdentifier(), outputFormat, lm)
```

src/main/java/org/lsc/Hooks.java
```diff
@@ -172,16 +156,14 @@ public final static void callHook(	String operationType,
...
 		try {
 			json = ow.writeValueAsString(lm.getLscAttributeModifications());
 		}
 		catch(Exception e) {
-			StringWriter sw = new StringWriter();
-			PrintWriter pw = new PrintWriter(sw);
-			e.printStackTrace(pw);
-			LOGGER.error("Error while encoding LSC modifications to json", sw.toString());
+			LOGGER.error("Error while encoding LSC modifications to json");
+			LOGGER.error("encoding error: ", e);
```
rouazana, line 166:
```suggestion
			LOGGER.error("Error while encoding LSC modifications to json", e);
```

src/main/java/org/lsc/Hooks.java
```diff
@@ -158,11 +141,12 @@ public final static void callHook(	String operationType,
...
-			PrintWriter pw = new PrintWriter(sw);
-			e.printStackTrace(pw);
-			LOGGER.error("Error while calling {} posthook {} with format {} for {}: {}",
-					operationType, hook, format, identifier, sw.toString());
+			LOGGER.error("Error while calling {} posthook {} with format {} for {}",
+					operationType,
+					hook,
+					outputFormat.toString(),
+					identifier);
+			LOGGER.error("posthook error: ", e);
```
rouazana, line 149:
```suggestion
			LOGGER.error("Error while calling {} posthook {} with format {} for {}",
					operationType,
					hook,
					outputFormat.toString(),
					identifier,
					e);
```
should work also

src/main/java/org/lsc/beans/syncoptions/ForceSyncOptions.java
```diff
@@ -128,28 +129,28 @@ public String getCondition(LscModificationType operation) {
...
-	public String getPostHookOutputFormat() {
-		return "";
+	public OutputFormat getPostHookOutputFormat() {
+		return OutputFormat.LDIF;
 	}
 
-	public String getCreatePostHook() {
-		return "";
+	public Optional<String> getCreatePostHook() {
+		return Optional.ofNullable(null);
```
rouazana, line 137:
```suggestion
		return Optional.empty();
```

src/main/java/org/lsc/beans/syncoptions/ForceSyncOptions.java
```diff
@@ -128,28 +129,28 @@ public String getCondition(LscModificationType operation) {
...
 	}
 
-	public String getDeletePostHook() {
-		return "";
+	public Optional<String> getDeletePostHook() {
+		return Optional.ofNullable(null);
```
rouazana, line 141:
```suggestion
		return Optional.empty();
```

src/main/java/org/lsc/beans/syncoptions/ForceSyncOptions.java
```diff
@@ -128,28 +129,28 @@ public String getCondition(LscModificationType operation) {
...
 	}
 
-	public String getUpdatePostHook() {
-		return "";
+	public Optional<String> getUpdatePostHook() {
+		return Optional.ofNullable(null);
```
rouazana, line 145:
```suggestion
		return Optional.empty();
```

src/main/java/org/lsc/beans/syncoptions/ForceSyncOptions.java
```diff
@@ -128,28 +129,28 @@ public String getCondition(LscModificationType operation) {
...
 	}
 
-	public String getChangeIdPostHook() {
-		return "";
+	public Optional<String> getChangeIdPostHook() {
+		return Optional.ofNullable(null);
```
rouazana, line 149:
```suggestion
		return Optional.empty();
```

src/main/java/org/lsc/beans/syncoptions/ForceSyncOptions.java
```diff
@@ -128,28 +129,28 @@ public String getCondition(LscModificationType operation) {
...
 	}
 
-	public String getPostHook(LscModificationType operation) {
-		return "";
+	public Optional<String> getPostHook(LscModificationType operation) {
+		return Optional.ofNullable(null);
```
rouazana, line 153:
```suggestion
		return Optional.empty();
```

src/main/java/org/lsc/beans/syncoptions/PropertiesBasedSyncOptions.java
```diff
@@ -197,43 +198,41 @@ public String getCondition(LscModificationType operation) {
...
-		return conf.getHooks().getChangeIdPostHook();
+	public Optional<String> getChangeIdPostHook() {
+		Optional<String> hook = Optional.ofNullable(conf.getHooks().getChangeIdPostHook()).filter(s -> !s.isEmpty());
+		return hook;
 	}
 
-	public String getPostHook(LscModificationType operation) {
-		String result = "";
+	public Optional<String> getPostHook(LscModificationType operation) {
+		Optional<String> result = Optional.ofNullable(null);
```
rouazana, line 235: you can simplify this method by removing the result method and returning directly into each case

src/main/java/org/lsc/runnable/CleanEntryRunner.java
```diff
@@ -110,9 +112,10 @@ public void run() {
 				// do it!
 				if (task.getDestinationService().apply(lm)) {
 					// Retrieve posthook for the current operation
-					String hook = syncOptions.getDeletePostHook();
-					String outputFormat = syncOptions.getPostHookOutputFormat();
-					Hooks.postSyncHook(hook, outputFormat, lm);
+					Optional<String> hook = syncOptions.getDeletePostHook();
+					OutputFormat outputFormat = syncOptions.getPostHookOutputFormat();
+					Hooks hookObject = new Hooks();
```
rouazana, line 117: hookObject (you can name it hooks) should be initialized in the constructor.

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -139,9 +141,10 @@ public boolean run(IBean entry) {
 			// if we got here, we have a modification to apply - let's do it!
 			if (task.getDestinationService().apply(lm)) {
 				// Retrieve posthook for the current operation
-				String hook = task.getSyncOptions().getPostHook(modificationType);
-				String outputFormat = task.getSyncOptions().getPostHookOutputFormat();
-				Hooks.postSyncHook(hook, outputFormat, lm);
+				Optional<String> hook = task.getSyncOptions().getPostHook(modificationType);
+				OutputFormat outputFormat = task.getSyncOptions().getPostHookOutputFormat();
+				Hooks hookObject = new Hooks();
```
rouazana, line 146: idem

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -139,9 +141,10 @@ public boolean run(IBean entry) {
...
+				hookObject.postSyncHook(hook, outputFormat, lm);
```
rouazana, line 147: also I think you can inline the parameters

rouazana (commented):
Sorry again a bunch of comments, but it starts to look really nice!

davidcoutadeur:
commit 45d8a326 is fixing the last bunch of comments.

rouazana (approved):

PR #243, opened 2023-10-30
Update izpack to 5.2.1

xavierba:
This fixes an issue when building with Java 14+
See https://openjdk.org/jeps/367

For whatever reason, `@{project.url}` is not expanded in ` src/install/install.xml]`, so it's commented out, but this needs to be fixed before this can be commited.


xavierba:
The issue with `@{project.url}` should be fixed now.

PR #245, opened 2023-10-31
Drop commas from cn in sample.csv

xavierba:


coudot:
I don't understand why you want to do this. The separator of fields is `;` so it is possible to have `,` in fields value.

xavierba:
sample.csv syntax is valid, but this leads to the following error when running lsc:
```
ERROR - Error while synchronizing ID cn=Clarke, Jonathan,ou=People,dc=example,dc=com: java.lang.RuntimeException: org.apache.directory.api.ldap.model.exception.LdapInvalidDnException: ERR_13605_START_AT_DESCR_EXPECTED Un
expected character ',' at position 20. Expected start of attributeType descr.
```

coudot:
You need to escape specific characters in DN.

See https://www.lsc-project.org/documentation/latest/upgrade.html#dn-escaping

```xml
<mainIdentifier>js:"cn=" + javax.naming.ldap.Rdn.escapeValue(srcBean.getDatasetFirstValueById("cn")) + ",OU=users,ou=demo,DC=example,DC=com"</mainIdentifier>

PR #246, opened 2023-11-06
RPM specfile cleanup

xavierba:
Extensive RPM specfile cleanup.

Closes #265

PR #247, opened 2023-11-06
Add systemd unitfiles 

### LINKED ISSUE #50: Make a systemd script

xavierba:
This applies on top of the RPM spec file cleanup merge request #246.
It fixes #50. This is different enough from what @davidcoutadeur did in #230, so I'd rather not try to improve #230 and submit a new one.
And indeed, thank you very much @davidcoutadeur, all your inputs were invaluable and I wouldn't have gone this far without you :-)


davidcoutadeur:
> This applies on top of the RPM spec file cleanup merge request #246. It fixes #50. This is different enough from what @davidcoutadeur did in #230, so I'd rather not try to improve #230 and submit a new one. And indeed, thank you very much @davidcoutadeur, all your inputs were invaluable and I wouldn't have gone this far without you :-)


Seems nice.
At first sight, we miss the equivalent for Debian, isn't it? (at least for new systemd service deployments, but ideally for all new changes brought in rpm specfile)

davidcoutadeur:
As discussed with @xavierba, this MR is obsolete and replaced by #260 

#260 only takes care about adding the systemd files. Integration in debian/red-hat will be done elsewhere.

PR #249, opened 2023-11-13
Use new ForgeRock repository URL

amottier:
During the build I got the following error:
```
Could not transfer artifact org.forgerock.ce.opendj:opendj-server:pom:2.6.4 from/to forgerock-community-repository (https://maven.forgerock.org/repo/community): status code: 401, reason phrase:  (401)
```

According to https://backstage.forgerock.com/knowledge/kb/article/a53108595, ForgeRock URL changed from https://maven.forgerock.org/repo/ to https://maven.forgerock.org/artifactory/.

This pull request updates the pom.xml to use thes new URL.

PR #250, opened 2023-11-14
Fix the JDK version range for jdk9 profile

amottier:
The version range was: `[9,11[`  leading to activation of `jdk9` profile when running on Java 11. This can be confirmed by running: `mvn help:active-profiles`. Actually `[9,11[` is not a valid range definition according to https://maven.apache.org/ref/3.9.5/maven-artifact/apidocs/org/apache/maven/artifact/versioning/VersionRange.html#createFromVersionSpec(java.lang.String) This change update the definition of version range to `[9,11)` that actually exclude Java 11 (`jdk11` will be activated in such situation).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -461,7 +461,7 @@
 		<profile>
 			<id>jdk9</id>
 			<activation>
-				<jdk>[9,11[</jdk>
+				<jdk>[9,11)</jdk>
 			</activation>
 			<properties>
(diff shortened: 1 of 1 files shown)


amottier:
Note that this PR requires #249 for the CI job to be successful.

rouazana (approved):
Thank you for the change and the explanation.

I let you rebase and we can merge if the pipeline is green.

amottier:
Rebase is done @rouazana PR should be ready to be merged.

PR #254, opened 2023-11-30
Upgrade logging libraries to last version

rouazana:


coudot:
Hello, will this replace #252 and #253 ?

coudot:
We have this error in CI:
```class file has wrong version 55.0, should be 52.0```

Any idea?

PR #260, opened 2024-01-08
Add a systemd script (#50)

### LINKED ISSUE #50: Make a systemd script

davidcoutadeur:
Fixes #50


davidcoutadeur:
This MR only takes care about adding the systemd files. Integration in debian / red-hat will be done elsewhere.

davidcoutadeur:
On more point to check in this PR: describe the way to duplicate the sync/async systemd script to have more than one task.
Possible leads:
- overload systemd script (systemctl edit)
- use systemd instances

We need to clarify this in documentation, and also define if admins should use /etc/default for LSC_CONFIG_DIR, LSC_SYNC_TASKS,...

davidcoutadeur:
> On more point to check in this PR: describe the way to duplicate the sync/async systemd script to have more than one task. Possible leads:
> 
>     * overload systemd script (systemctl edit)
> 
>     * use systemd instances
> 
> 
> We need to clarify this in documentation, and also define if admins should use /etc/default for LSC_CONFIG_DIR, LSC_SYNC_TASKS,...

We can only pass one parameter to a systemd instance.

But admins can still use systemd instances for defining the path to `/etc/default/lsc` to load.

davidcoutadeur:
I have provided an example of systemd instance in 527a318c

Not tested yet, but it should do the job. @xavierba if you have any remark? @coudot don't hesistate if you have an opinion on this topic.

davidcoutadeur:
systemd scripts seems ready in #260 
I am going to merge.

What I have done since last review:
- removed /etc/init.d/lsc support
- removed useless sed on /etc/default/lsc-* files (since these variables are now defined in corresponding systemd unit files)
- fix pid filename directly in systemd unit file
- removed `%systemd_preun lsc-async@.service` since systemd doesn't know which instance to stop (even with a joker)
- removed /etc/default/lsc (since not used any more)
- clean and simplify /etc/default/lsc-sync and /etc/default/lsc-async
- write the documentation (see https://github.com/lsc-project/documentation/blob/main/source/run.rst#using-systemd-service)
- test build + install + run + remove on debian12: OK
- test build + install +run + remove on rocky9: OK (warnings during preinstall, but not due to systemd)
- split long lines

PR #266, opened 2024-01-26
deb cleanup (#264)

### LINKED ISSUE #264: deb cleanup
This is the same issue as https://github.com/lsc-project/lsc/pull/246, but for debian packaging

davidcoutadeur:
Closes #264

### DIFF
--- a/src/install/debian/control
+++ b/src/install/debian/control
@@ -10,10 +10,15 @@ Vcs-Browser: https://github.com/lsc-project/lsc
 
 Package: lsc
 Architecture: all
-Depends: ${misc:Depends}
+Depends: cron, ${misc:Depends}
 Description: LDAP Synchronization Connector
  LSC offers a powerful transformation engine, based on a scripting language, 
  to easily manipulate data on the fly. Various identity management functions 
  are included for directory-specific compatibility - most notably Active Directory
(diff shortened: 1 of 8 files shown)


src/install/debian/control
```diff
@@ -10,10 +10,15 @@ Vcs-Browser: https://github.com/lsc-project/lsc
...
 Description: LDAP Synchronization Connector
  LSC offers a powerful transformation engine, based on a scripting language, 
  to easily manipulate data on the fly. Various identity management functions 
  are included for directory-specific compatibility - most notably Active Directory
  (changing passwords, account status, last logon, etc ...). LSC is an open source
  project written in Java, available under the BSD license.
+
+Package: nagios-plugins-lsc
+Architecture: all
+Depends: perl:any
```
xavierba, line 23: I think some perl dependencies are missing here, namely `File::Basename`, `Getopt::Long` and `strict`.

It seems the proper way to declare this in Debian is to use `Depends: ${perl:Depends}` which will then use automatic dependencies detection. This in turn requires `Build-Depends: debhelper (>= 3.0.18)`.

See https://www.debian.org/doc//packaging-manuals/perl-policy/ch-module_packages.html#s-dh-perl

PR #267, opened 2024-02-12
Return unicodePwd as byte[] instead of String

coudot:
Simple patch to manage AD unicodePwd encoding issue

This also requires to set unicodePwd as binary attribute in the LDAP connection


artlog:
this is a fine shortest fix. +1 !

rouazana (approved):

PR #269, opened 2024-03-12
fix: sample code doesn't run from Debian package

### LINKED ISSUE #268: hsqldb related samples from Debian package don't find hsqldb*.jar
Running `/usr/share/doc/lsc/hsqldb/bin/lsc-sample` from the Debian package raises the following message:

```
ls: cannot access '/usr/share/doc/lsc/hsqldb/bin/../../../lib/hsqldb*.jar': No such file or directory
```

When variable is properly set, the startup crashes:

```
Error: Could not find or load main class org.hsqldb.Server
```

pipoprods:
Fixes #268


coudot:
Thanks @pipoprods !

PR #272, opened 2024-04-04
Draft: Add java 17 in CI

xavierba:
This does include the update to izpack from #243, which is needed for java 14+.
This is a draft, as I just want to check the CI with java 17 (and later with java 21).


xavierba:
Superseded by #303

PR #259, opened 2023-12-21
Optimization for multi-valued LDAP attributes such as groups uniqueMember

rouazana:
See #255

Issue #255, opened 2023-12-08
Optimization for multi-valued LDAP attributes such as groups uniqueMember


FlagonC:
When configuring LSC synchronization, it is currently not possible to optimize modifications to be able to make add and delete modifications for multivalued attributes containing many entries, such as groups with uniqueMember.
This can lead to slowness when used with attributes containing thousands of entries.

For exemple, i have a test group contenning multiples users:

> dn: cn=test,ou=group,dc=test
> uniqueMember: uid=test01,ou=people,dc=test
> uniqueMember: uid=test02,ou=people,dc=test
> uniqueMember: uid=test03,ou=people,dc=test

If i add the user test04, LSC will make the modification like this:

> dn: cn=test,ou=group,dc=test
> changetype: modify
> replace: uniqueMember
> uniqueMember: uid=test01,ou=people,dc=test
> uniqueMember: uid=test02,ou=people,dc=test
> uniqueMember: uid=test03,ou=people,dc=test
> uniqueMember: uid=test04,ou=people,dc=test

The most optimized solution would be to do the following for large groups 

> dn: cn=test,ou=group,dc=test
> changetype: modify
> add: uniqueMember
> uniqueMember: uid=test04,ou=people,dc=test

[https://github.com/lsc-project/lsc/issues/70](url)
We can do add by following this function, but not for delete

LiquidFenrir:
Hello @coudot, I have fixes/improvements for this implemented but not uploaded/PR'd yet, I'd like to know which one would be preferred as there are two, each with pros and cons.

1. "optimizing" the FORCE policyType in the case of a REPLACE_VALUES operation in BeanComparator.java to use smaller ADD and DELETE modifications instead of a complete REPLACE.
Leverages SetUtils.findMissingNeedles in the opposite direction of MERGE to find the "excess" elements in the destination compared to the source.

2. Involves adding two policyTypes in the XSD schema, DIFFERENCE and KEEP_SOURCE (names unimportant, but tries to make the meaning obvious), also in case of a REPLACE_VALUES operation:
    - DIFFERENCE is the set difference operation: output = destination - source. Applying the DELETE modification to this output element set, makes this the opposite of the MERGE policyType: it removes excess elements in the destination (whereas MERGE adds the missing elements in the destination).
[...]

PR #284, opened 2024-05-10
Optimization for multi-valued LDAP attributes such as groups uniqueMember

intmgroupe:
See #255.
This pull request replaces #259 as per the discussion in its thread.

The goal is to make FORCE operations on large datasets faster by doing a few `add` and `delete` instead of one constantly huge `replace` when it is beneficial to do so.
The testing on this is nowhere near substantial, nor are any test cases updated to reflect the change in behaviour.

These changes were developed by INTM on behalf of EDF, who noticed the slowdown in real use, and the fix is succesfully working for them.

Please let us know if something is wrong or doesn't fit the guidelines adopted by the LSC project, so we can make any required modification before this can be merged.

### DIFF
--- a/src/main/java/org/lsc/beans/BeanComparator.java
+++ b/src/main/java/org/lsc/beans/BeanComparator.java
@@ -290,6 +290,7 @@ public final class BeanComparator {
 			LscDatasetModificationType operationType = getRequiredOperationForAttribute(toSetAttrValues, dstAttrValues);
 
 			// Build the modification
+			List<LscDatasetModification> multiMi = null;
 			LscDatasetModification mi = null;
 			switch (operationType) {
 				case DELETE_VALUES:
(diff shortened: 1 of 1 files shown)


davidcoutadeur:
Giving a look to this PR.

src/main/java/org/lsc/beans/BeanComparator.java
```diff
@@ -320,10 +321,32 @@ private static LscModifications getUpdatedObject(
 
 				case REPLACE_VALUES:
 					if (attrStatus == PolicyType.FORCE) {
-						if (isModified(dstBean, dstAttrValues, toSetAttrValues)) {
+						// check if there are any extra values to be added
+						Set<Object> missingValues = SetUtils.findMissingNeedles(dstAttrValues, toSetAttrValues);
+						// check if there are any extra values to be removed
+						Set<Object> extraValues = SetUtils.findMissingNeedles(toSetAttrValues, dstAttrValues);
+
+						if((missingValues.size() + extraValues.size()) >= toSetAttrValues.size()) {
```
davidcoutadeur, line 329: I am not sure at which point one replace is more efficient than 1 add + 1 delete.

Maybe it can be a configuration parameter? The condition would be:

```
(missingValues.size() + extraValues.size()) >= max_value_modifications
```

It would make the LSC modifications more predictable.

However I understand that depending on the entry size has also its benefits. For example, with the previous condition, if we had an entry with 1000 values, and max_value_modifications = 100, then 50 adds and 51 adds would lead to a big replace, which may be less efficient than adding 50 values + removing 51 values.

Without further efficiency analysis, I think we can keep the initial proposition in this PR.

davidcoutadeur (commented):
I like this PR, it is quite elegant in my opinion.

I have only 2 general comments:
- it would be really nice to have a corresponding test. Maybe we could grab the LSC log showing that we are in the condition: `(missingValues.size() + extraValues.size()) >= toSetAttrValues.size()`?
- as it is a behavior change for everyone, I think we must document this (in main documentation + in release notes)

davidcoutadeur:
> I like this PR, it is quite elegant in my opinion.
> 
> I have only 2 general comments:
> 
>     * it would be really nice to have a corresponding test. Maybe we could grab the LSC log showing that we are in the condition: `(missingValues.size() + extraValues.size()) >= toSetAttrValues.size()`?
> 
>     * as it is a behavior change for everyone, I think we must document this (in main documentation + in release notes)

I have just written the corresponding unit test demonstrating the correct selection of 1 big replace / 1 add + 1 delete.

See this PR: https://github.com/lsc-project/lsc/pull/285

davidcoutadeur:
> I like this PR, it is quite elegant in my opinion.
> 
> I have only 2 general comments:
> 
>     * it would be really nice to have a corresponding test. Maybe we could grab the LSC log showing that we are in the condition: `(missingValues.size() + extraValues.size()) >= toSetAttrValues.size()`?
> 
>     * as it is a behavior change for everyone, I think we must document this (in main documentation + in release notes)

I have also written the documentation and upgrade notes in corresponding project, see: https://github.com/lsc-project/documentation/issues/5

davidcoutadeur:
For me, everything is ready for merging. @rouazana and @soisik, I don't know if you have any more remarks?

davidcoutadeur:
This PR is replaced by #285

PR #289, opened 2024-05-21
Fix lsc-async unitfile and a bit more

xavierba:
@davidcoutadeur This should polish up the systemd work we started quite some time ago.


davidcoutadeur:
Linked to #50

Issue #280, opened 2024-04-17
Java 21: GraalJSEngineFactory could not be instantiated


xavierba:
CI fails when using Java 21 with the following error:

```
ScriptEngineManager providers.next(): javax.script.ScriptEngineFactory: Provider com.oracle.truffle.js.scriptengine.GraalJSEngineFactory could not be instantiated
ScriptEngineManager providers.next(): javax.script.ScriptEngineFactory: Provider com.oracle.truffle.js.scriptengine.GraalJSEngineFactory could not be instantiated
Error:  Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.645 s <<< FAILURE! - in org.lsc.beans.syncoptions.PropertiesBasedSyncOptionsTest
Error:  org.lsc.beans.syncoptions.PropertiesBasedSyncOptionsTest.testJS  Time elapsed: 0.606 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "org.lsc.utils.ScriptableEvaluator.evalToObjectList(org.lsc.Task, String, java.util.Map)" because "se" is null
	at org.lsc.utils.ScriptingEvaluator.evalToObjectList(ScriptingEvaluator.java:159)
	at org.lsc.beans.syncoptions.PropertiesBasedSyncOptionsTest.testJS(PropertiesBasedSyncOptionsTest.java:164)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:377)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:284)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:248)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:167)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:456)
[...]

PR #291, opened 2024-06-04
fallback to rhino if no default javascript is found (#280)

artlog:
- avoid NullPointerException
- always default to a non null defaut scripting implementation
  - default to ErrorMissingScriptEvalauator to fail a first unsupported scripting usage
- try to uee rhino if no js is


artlog:
to fix : incorrect defaut if no graaljs and no rhino ( defautImplementation can be reset to null ).

Issue #176, opened 2021-10-22
list of DN elements is truncated when using search method on JndiService


asouppart:
The number of elements returned by the search method is limited by MaxPageSize configuration of the ldap server.
Even if the PageSize configuration is populated with a positive number.

coudot:
Seems the issue is in the scripting engine (`ldap.search()`)

coudot:
@soisik @davidcoutadeur @rouazana we would like a review on #299

coudot:
Global OpenLDAP setting to get a soft limit of 100 entries and hard limit unlimited:
```
dn: cn=cconfig
olcSizeLimit: size.soft=100 size.hard=unlimited
```

We can test the pagedResult control with ldapsearch like this:
```
ldapsearch -x -H ldap://localhost -D uid=coudot,ou=users,dc=example,dc=com -w secret -b dc=example,dc=com -E pr=100/noprompt
```

davidcoutadeur:
I have reviewed part of the code, and done some tests about this issue.

Here is the status:
- first, it must be noted that the hard size limit cannot be exceeded. OpenLDAP documentation states:

```
If the LDAP client adds the pagedResultsControl to the search operation, the hard size limit is used by default, because the request for a specific page size is considered an explicit request for a limitation on the number of entries to be returned. However, the size limit applies to the total count of entries returned within the search, and not to a single page.

Additional size limits may be enforced for paged searches.
```

- we must check that the AD limitation can also be bypassed by the paged control.

- when running LSC with a paged search of 3 entries, and OpenLDAP configured with: `olcLimits: {1}group/groupOfNames/member="cn=adminlimit,ou=groups,dc=my-domain,dc=com" size.soft=3 size.hard=unlimited time=unlimited`, LSC can indeed get all entries (5 entries) when doing a srcLdap.search, but I get some errors, maybe not linked to this issue:

```
sept. 25 18:50:38 - ERROR - Error while modifying entry uid=test,ou=people2,dc=my-domain,dc=com in directory :javax.naming.OperationNotSupportedException: [LDAP: error code 12 - critical extension is unavailable]; remaining name 'uid=test,ou=people2'
sept. 25 18:50:38 - ERROR - Error while synchronizing ID uid=test,ou=people2,dc=my-domain,dc=com: java.lang.Exception: Technical problem while applying modifications to the destination
```

at OpenLDAP side:

```
[...]

PR #299, opened 2024-07-26
Fixed bug override globally configured pageSize variable in the middle of the code

abpai94:
#176

### DIFF
--- a/src/main/java/org/lsc/jndi/JndiServices.java
+++ b/src/main/java/org/lsc/jndi/JndiServices.java
@@ -155,6 +155,9 @@ public final class JndiServices {
 	
 	/** Remember connection properties to reconnect */
 	private Properties connProps;
+
+	/** Remember default resquest controls */
+	private Control[] defaultRequestControls;
 	
 	/**
 	 * Initiate the object and the connection according to the properties.
@@ -233,15 +236,6 @@ public final class JndiServices {
 			throw new NamingException(e.getMessage());
 		}
 
-		String pageSizeStr = (String) ctx.getEnvironment().get("java.naming.ldap.pageSize");
-		if (pageSizeStr != null) {
-			pageSize = Integer.parseInt(pageSizeStr);
-		} else {
-			pageSize = -1;
-		}
-
-		sortedBy = (String) ctx.getEnvironment().get("java.naming.ldap.sortedBy");
-		
 		String recursiveDeleteStr = (String) ctx.getEnvironment().get("java.naming.recursivedelete");
 		if (recursiveDeleteStr != null) {
 			recursiveDelete = Boolean.parseBoolean(recursiveDeleteStr);
@@ -741,28 +735,37 @@ public final class JndiServices {
 	private List<String> doGetDnList(final String base, final String filter,
 			final int scope) throws NamingException {
 		NamingEnumeration<SearchResult> ne = null;
-		List<String> iist = new ArrayList<String>();
+		List<String> list = new ArrayList<String>();
 		try {
+			contextRequestControls();
 			SearchControls sc = new SearchControls();
 			sc.setDerefLinkFlag(false);
 			sc.setReturningAttributes(new String[]{"1.1"});
 			sc.setSearchScope(scope);
 			sc.setReturningObjFlag(true);
-			ne = ctx.search(base, filter, sc);
-			
-			String completedBaseDn = "";
-			if (base.length() > 0) {
-				completedBaseDn = "," + base;
-			}
+			byte[] pagedResultsResponse = null;
+			do {
+				ne = ctx.search(base, filter, sc);
+				String completedBaseDn = "";
+				if (base.length() > 0) {
+					completedBaseDn = "," + base;
+				}
 			while (ne.hasMoreElements()) {
-				iist.add(((SearchResult) ne.next()).getName() + completedBaseDn);
+				list.add(ne.next().getName() + completedBaseDn);
 			}
+			pagedResultsResponse = pagination();
+			} while (pagedResultsResponse != null);
 		} catch (NamingException e) {
 			LOGGER.error(e.toString());
 			LOGGER.debug(e.toString(), e);
 			throw e;
+		} catch (IOException e) {
+			LOGGER.error(e.toString());
+			LOGGER.debug(e.toString(), e);
+		} finally {
+			ctx.setRequestControls(defaultRequestControls);
 		}
-		return iist;
+		return list;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -1137,69 +1139,25 @@ public Map<String, LscDatasets> doGetAttrsList(final String base,
...
-			
-			if(sortedBy != null) {
-			    extControls.add(new SortControl(sortedBy, Control.CRITICAL));
-			}
-
-			if (extControls.size() > 0) {
-				ctx.setRequestControls(extControls.toArray(new Control[extControls.size()]));
-			}
-
-			byte[] pagedResultsResponse = null;
```
davidcoutadeur, line 1161: I have no time to test this, but the previous lines adding extended controls seem quite useful, aren't they?
Why did you remove them?

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -1213,11 +1171,6 @@ public Map<String, LscDatasets> doGetAttrsList(final String base,
 			LOGGER.error(e.toString());
 			LOGGER.debug(e.toString(), e);
 			
-		} catch (IOException e) {
-			// clear requestControls for future use of the JNDI context
-			ctx.setRequestControls(null);
-			LOGGER.error(e.toString());
-			LOGGER.debug(e.toString(), e);
```
davidcoutadeur, line 1220: It also seems useful to log the catched errors, isn't it?

src/main/java/org/lsc/jndi/JndiServices.java
abpai94: I will re-introduce the sortedBy code in the constructor. It appears to be global configuration for each request that is made to LDAP regardless of via script or otherwise.

abpai94:
@davidcoutadeur I have understood the problem better and proposed a new fix. The commits have been squashed as requested.

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -745,28 +739,35 @@ public List<String> getDnList(final String base, final String filter,
...
+				String completedBaseDn = "";
+				if (base.length() > 0) {
+					completedBaseDn = "," + base;
+				}
+				while (ne.hasMoreElements()) {
+					list.add(ne.next().getName() + completedBaseDn);
+				}
+				pagedResultsResponse = pagination();
+			} while (pagedResultsResponse != null);
+			contextRequestControls();
```
davidcoutadeur, line 761: this one seems useless

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -745,28 +739,35 @@ public List<String> getDnList(final String base, final String filter,
...
 		} catch (NamingException e) {
 			LOGGER.error(e.toString());
```
davidcoutadeur, line 763: in every catch structure, you should cal the clearsearchcontrols function to ensure the controls are removed (see my comment below for more details about this function)

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -1135,30 +1136,8 @@ public Map<String, LscDatasets> doGetAttrsList(final String base,
 		constraints.setReturningAttributes(attributes);
 		constraints.setSearchScope(scope);
 		constraints.setReturningObjFlag(true);
-
```
davidcoutadeur, line 1138: Same comment as previous:

In this function, you should:
* call a new function setsearchcontrols() before the search. This function would set the pagesize and sortedby controls
* call a new function clearsearchcontrols() after the search. This function would get the current controls (with  getRequestControls), and remove only the paged and sortedby controls if they exist. And also take care to call the clearsearchcontrols in the catch clauses.

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -1180,48 +1159,74 @@ public Map<String, LscDatasets> doGetAttrsList(final String base,
...
+				requestControls.add(new PagedResultsControl(pageSize, Control.CRITICAL));
+				ctx.setRequestControls(requestControls.toArray(new Control[requestControls.size()]));
+			}
+
+			// Setting global sortedBy variable
+			sortedBy = (String) ctx.getEnvironment().get("java.naming.ldap.sortedBy");
+			if (sortedBy != null) {
+				List<SortControl> requestControls = new ArrayList<>();
+				requestControls.add(new SortControl(sortedBy, Control.CRITICAL));
+				ctx.setRequestControls(requestControls.toArray(new Control[requestControls.size()]));
```
davidcoutadeur, line 1223: Here, I think you should prepare a global requestControls list, and call setRequestControls only once.

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -1180,48 +1159,74 @@ public Map<String, LscDatasets> doGetAttrsList(final String base,
...
+		if (respCtls != null) {
+			for(Control respCtl : respCtls) {
+				if (respCtl instanceof PagedResultsResponseControl) {
+					pagedResultsResponse = ((PagedResultsResponseControl) respCtl).getCookie();
+				}
+			}
+		}
+		if (pagedResultsResponse != null) {
+			ctx.setRequestControls(new Control[]{
+					new PagedResultsControl(pageSize, pagedResultsResponse, Control.CRITICAL)});
```
davidcoutadeur, line 1199: I am not sure, but here maybe you should get the previous controls, remove any previous pagedcontrol, and add the new pagedcontrol with the cookie.

davidcoutadeur (changes requested):
Here is my complete review

src/main/java/org/lsc/jndi/JndiServices.java
```diff
@@ -745,28 +739,35 @@ public List<String> getDnList(final String base, final String filter,
...
-			ne = ctx.search(base, filter, sc);
-			
-			String completedBaseDn = "";
-			if (base.length() > 0) {
-				completedBaseDn = "," + base;
-			}
-			while (ne.hasMoreElements()) {
-				iist.add(((SearchResult) ne.next()).getName() + completedBaseDn);
-			}
+			byte[] pagedResultsResponse = null;
```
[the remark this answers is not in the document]
abpai94: Fixed!

src/main/java/org/lsc/jndi/JndiServices.java
abpai94: Using the try-catch-finally block to handle clearing the request controls.

src/main/java/org/lsc/jndi/JndiServices.java
abpai94: As this code was duplicated in multiple location pagination() is called instead.

src/main/java/org/lsc/jndi/JndiServices.java
abpai94: Fixed

PR #53, opened 2018-04-30
Supervision logs multitasks

floriancaillet:
In our LSC practices, we realised a multi-tasks synchronization process.
We needed to supervise this process, but status log and perl script check_lsc_status don't work with  multi-tasks synchronization .
So, we add to status logs, informations about taskname and taskmode (changing AbstractSynchronize.java).
And we have made a new perl script for multi-tasks status log file analysis.

We think this changes could be integrated in v2.1.
What do you think about that ?

Best regards,
Florian CAILLET

PR #300, opened 2024-08-19
Log task name+mode as well

brammeleman:
As discussed here: https://mail.ow2.org/wws/arc/lsc-users/2024-07/msg00006.html

Based on the pull request that can't be pulled anymore: https://github.com/lsc-project/lsc/pull/53

It builds, but I can't find out how to run the freshly made lsc. It complains about a manifest:
```
➜ java -jar target/lsc-core-2.2-SNAPSHOT.jar 
no main manifest attribute, in target/lsc-core-2.2-SNAPSHOT.jar
```

I left out modifications to the perl file as I can't test the code.

### DIFF
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -1,7 +1,7 @@
 Build distribution
 ==================
 
-$ mvn -Popendj --file pom.xml clean package
+$ mvn -Popendj clean package
 
 tar.gz and .zip files available in target/ directory
 
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -1,7 +1,7 @@
 Build distribution
 ==================
 
-$ mvn -Popendj clean package
+$ mvn -Popendj --file pom.xml clean package
 
 tar.gz and .zip files available in target/ directory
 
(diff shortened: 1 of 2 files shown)


BUILD.txt
```diff
@@ -1,7 +1,7 @@
 Build distribution
 ==================
 
-$ mvn -Popendj clean package
+$ mvn -Popendj --file pom.xml clean package
```
brammeleman, line 4:
```suggestion
$ mvn -Popendj clean package
```

src/main/java/org/lsc/AbstractSynchronize.java
```diff
@@ -231,7 +231,7 @@ protected final boolean synchronize2Ldap(final Task task) {
 			LOGGER.info("If you want to avoid this message, " + "increase the time limit by using dedicated parameter.");
 		}
 
-		logStatus(counter);
+		logStatus(task.getName(), Task.Mode.clean.toString(), counter);
```
coudot, line 234: There is an issue here that is fixed by:
```suggestion
		logStatus(task.getName(), Task.Mode.sync.toString(), counter);
```

coudot:
I merge this into a work branch to add the modification of Perl script

PR #301, opened 2024-08-27
Provide a cleaner lsc.xml sample file

coudot:
The goal is to have a default lsc.xml that is easy to understand.

Here I choose to have a simple task to synchronize users between 2 LDAP directories, with some interesting features of LSC:
* Have a different pivot attribute between source and destination (uid/sAMAccountName)
* Escape RDN value
* Force objectClass values (top/user)
* Set a static value in an attribute (description)
* Map attributes (uid/sAMAccountName)
* Automatically synchronize attributes (cn, givenName, mail, sn)


davidcoutadeur:
Seems great! Thanks

PR #305, opened 2024-09-04
Fix: Added a function to improve jscript prefix identifier

### LINKED ISSUE #360: improve jscript prefix identifier
This issue for discussing the PR #305 made by @abpai94

abpai94:
When using the following prefix to scripts such as `rjs` or others:
```
<dataset>
    <name>member</name>
    <policy>FORCE</policy>
    <forceValues>
        <string>
              <![CDATA[rjs:
              ...jscript_function...
              ]]>
        </string>
    </forceValues>
</dataset>
```
The following error is thrown:
```
java.lang.NullPointerException: Cannot invoke "org.lsc.utils.ScriptableEvaluator.evalToObjectList(org.lsc.Task, String, java.util.Map)" because "se" is null
```
This occurs due to the prefix containing the following:
```
\n                            
            rjs
``` 
The `.contains` method in Maps object is unable to match this pattern.

This functionality works well with the following script:
```
<mainIdentifier>rjs:...jscript_function...</mainIdentifier>
```
As the script is done in-line the `.contains` method is able to match and find the appropriate jscript evaluator.

I have added a new function that uses pattern matching with the help of a small regex function which should resolve the above described issue.

I have tested this in scenarios where no script engine has been defined without posing any issue.
Also I have built on top of the work in #280 to seamlessly merge this minor fix.

Closes #360


abpai94:
@philhaworteks I have built this minor fix on top of your work in #280. We can merge when your changes are approved.

artlog:
This is not working with java 8 due to  java.util.Optional.or that was provided only from java 9.
java.lang.NoSuchMethodError: java.util.Optional.or(Ljava/util/function/Supplier;)Ljava/util/Optional;

abpai94:
I used java 9 to get the Optional code to work. I changed the code to the following:
```
			defaultImplementation = Optional.ofNullable(Optional.ofNullable(instancesTypeCache.get("js"))
                    .orElse(instancesTypeCache.get("rjs")));
```
It doesn't look pretty but it works.
It ensures that an optional is provided with `rjs` script evaluator.

rouazana:
I'm not sure this is worth it as this solution works fine:
```
<string><![CDATA[rjs:
              ...jscript_function...
              ]]></string>
```

abpai94:
Its a minor change that added additional tolerance to detect inline javascript. Seems much cleaner to use regex to determine the JS engine from the prefix.

I made additional changes in the most recent commit that adds further tolerance to detect a variety of cases.
Expected use case:
```
        <mainIdentifier><![CDATA[js:
           ...jscript...
          ]]></mainIdentifier>
```
Other use case 1:
```
        <mainIdentifier><![CDATA[
           js:
           ...jscript...
          ]]></mainIdentifier>
```
Other use case 2:
```
        <mainIdentifier>
          <![CDATA[js:
           ...jscript...
          ]]></mainIdentifier>
```
Other use case 3:
```
        <mainIdentifier>
          <![CDATA[
          js:
           ...jscript...
          ]]></mainIdentifier>
```
Other use case 4:
```
        <mainIdentifier><![CDATA[
           
           js:
           ...jscript...
          ]]></mainIdentifier>
```
Or any other variants that might be occur.

abpai94:
@rouazana If this seems unnecessary feel free to close off the issue. Found the lack of JS engine detection a minor inconvenience personally.

src/main/java/org/lsc/utils/ScriptingEvaluator.java
```diff
@@ -115,13 +116,33 @@ public static void contribute(String implementationName,
...
 
+	/**
+	 * Matches the prefix specifying the jscript evaluator.
+	 *
+	 * @param jscript the prefix
+	 * @return the matched jscript evaluator
+	 */
+	public String matchJScriptEvaluator(String jscript) {
+		for (String jscriptEngine : instancesTypeCache.keySet()) {
+			Pattern pattern = Pattern.compile("((\\n.*)+)" + jscriptEngine);
```
rouazana, line 134: As far as I know it's recommended to compile patterns only once, as it can be a heavy computation.

rouazana, line 134: I fear this pattern matches too many things, for example:
```
js:java.lang.System.out.println("rjs")
```
could match rjs
But also:
```
rjs:true
```
could match js

Could you add some tests to ensure your pattern matches exactly what you mean?

rouazana (commented):
I'm not against it as it can improve this piece of code, but I guess there are currently some issues.

src/main/java/org/lsc/utils/ScriptingEvaluator.java
abpai94: Created a method to pre-compile all regex pattern match combinations.

abpai94: Added a few unit tests for the usecases you have highlighted.

abpai94:
@rouazana I have implemented some tests and ameliorated the pattern compilation to reduce the overhead.

rouazana (commented):
Thank you very much for taking my comments into account!

I have only 2 remaining suggestions:
1. Try to name your tests with a meaningful function name, for example `scriptShouldEvaluateWhenStartingWithANewLine` for `testJsPrefix1`
2. I'm not sure you are really testing what I mean, for example that the *right* JS engine is chosen if the `js:` or `rjs:` strings appear in different places in the expression.

davidcoutadeur:
I close this PR as it has many conflicts and has been rewritten in #361

PR #324, opened 2025-01-10
fix slow performance on graalvm with jdk11 (#322)

### LINKED ISSUE #322: Warning : slow performance on GraalVM with JDK11
When compiling and running on JDK11, I get multiple warnings:

```
[engine] WARNING: The polyglot context is using an implementation that does not support runtime compilation.
The guest application code will therefore be executed in interpreted mode only.
Execution only in interpreted mode will strongly impact the guest application performance.
For more information on using GraalVM see https://www.graalvm.org/java/quickstart/.
To disable this warning the '--engine.WarnInterpreterOnly=false' option or use the '-Dpolyglot.engine.WarnInterpreterOnly=false' system property.
```

This is probably due to bad/missing dependencies or packaging in pom.xml.

davidcoutadeur:
Closes #322

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,7 @@
 		<staging.dir>${project.build.directory}/staging</staging.dir>
 		<jmockit-version>1.5</jmockit-version>
 		<graalvm.version>22.3.5</graalvm.version>
+		<graaljs.version>24.1.1</graaljs.version>
 		<compiler.dir>${project.build.directory}/compiler</compiler.dir>
 	</properties>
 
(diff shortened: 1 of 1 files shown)


davidcoutadeur:
This may require additional cleaning on graalvm declarations in pom.xml:
- use the same variable for graalvm.version and graaljs.version?
- get rid of useless dependencies? (there seem to be an a refactoring of graalvm sdk, see: https://www.graalvm.org/release-notes/JDK_21/)
- check if graalvm* artifactItems in maven-dependency-plugin plugin are still necessary?

PR #335, opened 2025-02-10
Ele junit5 ads

elecharny:
A complete Junit5 migration, with OpenDJ removal and jmockit replacement by Mockito.


elecharny:
Wrong ApacheDS version was used.

Issue #326, opened 2025-01-13
Tests failure when building with Java 11


elecharny:
Here is the build environment I was using:

```
$ mvn --version
OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
OpenJDK 64-Bit Server VM warning: Ignoring option MaxPermSize; support was removed in 8.0
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /opt/apache-maven-3.9.6
Java version: 11.0.11, vendor: AdoptOpenJDK, runtime: /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
Default locale: fr_FR, platform encoding: UTF-8
OS name: "mac os x", version: "11.1", arch: "x86_64", family: "mac"
```

Running the ```$ mvn -Popendj test``` command, I get:

```
[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 0.924 s <<< FAILURE! - in org.lsc.Ldap2LdapHookSyncTest
[ERROR] org.lsc.Ldap2LdapHookSyncTest.testLdap2LdapLDIFHookSyncTest  Time elapsed: 0.429 s  <<< FAILURE!
java.lang.AssertionError: Error while reading hook-ldif-create.log
	at org.lsc.Ldap2LdapHookSyncTest.checkLDIFSyncResults(Ldap2LdapHookSyncTest.java:173)
	at org.lsc.Ldap2LdapHookSyncTest.testLdap2LdapLDIFHookSyncTest(Ldap2LdapHookSyncTest.java:142)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #336, opened 2025-02-10
Ele junit5 ads AM27

### LINKED ISSUE #330: Switch to Junit5, Mockito, ApacheDS, Java 11/17/21
All the existing tests are based on ```Junit 4```. This is fine, as soon as you are running them on an antiquated version of **Eclipse**.
Recent version of **Eclipse** default on Junit5, and you have to configure each single test to execute with ```Junit4``` if you want to debug them, which is extra painful...

elecharny:
Switched back to ApacheDS AM27

Closes #330


elecharny:
Removed Windows RC from files (https://github.com/lsc-project/lsc/issues/339)
Replaced spaces by tabs in files

elecharny:
I'll stop working on this branch which contains too many changes related to various tickets.

It fixes:
* https://github.com/lsc-project/lsc/issues/339
* https://github.com/lsc-project/lsc/issues/333
* https://github.com/lsc-project/lsc/issues/330
* https://github.com/lsc-project/lsc/issues/326

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -155,31 +153,40 @@ public final void testLdap2LdapLDIFHookSyncTest() throws Exception {
...
 		try {
 			File hookFile = new File("hook-ldif-" + operation + ".log");
-			Scanner hookReader = new Scanner(hookFile);
 
-			while (hookReader.hasNextLine()) {
-				String data = hookReader.nextLine();
-				hookResults.add(data);
-			}
-			hookReader.close();
+            Thread.sleep(1000L);
```
davidcoutadeur: Is this mandatory to have Thread.sleep for this test to work?
If so, the test design is a little poor: it would break on systems with a slow disk. But I have currently no better idea.

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -155,31 +153,40 @@ public final void testLdap2LdapLDIFHookSyncTest() throws Exception {
...
+
+            if (!hookFile.exists()) {
+                System.out.println("!!!!!!!!!!!!!!!!!!! File " + hookFile.getAbsolutePath() + " does not exist");
+            }
```
davidcoutadeur: Not sure this is wanted in the test, is it?

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -193,23 +200,32 @@ private final void checkJSONSyncResults(String operation, JsonNode expectedEntry
 		List<String> hookResults = new ArrayList<String>();
 		try {
 			File hookFile = new File("hook-json-" + operation + ".log");
-			Scanner hookReader = new Scanner(hookFile);
+            
+            Thread.sleep( 1000L );
```
davidcoutadeur: same comment as previous

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
```diff
@@ -193,23 +200,32 @@ private final void checkJSONSyncResults(String operation, JsonNode expectedEntry
...
+            
+            if (!hookFile.exists()) {
+                System.out.println( "!!!!!!!!!!!!!!!!!!! File " + hookFile.getAbsolutePath() + " does not exist" );
+            }
```
davidcoutadeur: same comment as previous

davidcoutadeur (changes requested):
Some comments and minor demands about the tests

src/test/java/org/lsc/beans/BeanComparatorTestOld.java
```diff
@@ -0,0 +1,558 @@
+/*
```
davidcoutadeur: What's the point of having the file BeanComparatorTestOld.java? Is this a backup of BeanComparatorTest.java?

If so, we should remove it

src/test/java/org/lsc/jmx/LscJmxTest.java
```diff
@@ -74,6 +75,7 @@
  * This test covers all the JMX capabilities
  * @author Sebastien Bahloul &lt;seb@lsc-project.org&gt;
  */
+@Disabled
```
davidcoutadeur: Why is it disabled? Is JMX not supported any more?

If JMX is still supported, I think we should have another test instead of just disabling it.

davidcoutadeur (changes requested):
Some comments and demands about the ApacheDS + mockito commit

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
elecharny: No, I think it should be removed

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
elecharny: indeed ;-)

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
elecharny: same answer

src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
elecharny: same

pom.xml
```diff
@@ -70,8 +75,8 @@
 		<!--Force UTF-8 for the build -->
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<staging.dir>${project.build.directory}/staging</staging.dir>
-		<jmockit-version>1.5</jmockit-version>
-		<graalvm.version>22.3.5</graalvm.version>
+		<jmockit-version>1.7</jmockit-version>
```
davidcoutadeur, line 78: Sorry, but I have pushed to master the PR #324, which introduces a conflict. Can I let you handle the conflict? Don't hesitate to ask if you have any doubt about the resolution.

davidcoutadeur (changes requested):

elecharny:
I have merged the changes. The build passes.

elecharny:
Hi Raphael,

when I started working - and even as a student -, almost 40 years ago, I was all in for tabs for many reasons, one of them being that is saved 'space' (one tab, one char when you eat 4 bytes when using spaces). But back then, you only had 20Mb hard drive, 160Kb floppy disks, and size was of the essence.
Nowadays, the main point about spaces vs tabs is that when it comes to enforce the format choice, it's freaking easy to find tabs in a file using grep, and replace them with spaces. Doing the opposite is a nightmare.
And another aspect is that you want to be able to chose the indentation size: 4 spaces for code, 2 spaces for XML, for instance. Which is hard to do with tabs.

davidcoutadeur:
final rebased PR in #344

PR #349, opened 2025-02-20
Simplify the JndiServices.contextRequestControls method

elecharny:
o Removed the private sortedBy field, it has no reason to exist 
o Minor code reformating
o Merged the two controls addition into one operation


elecharny:
Fix for issue #350

PR #374, opened 2025-03-05
369 db2ldap

elecharny:
Close 369: Added a test for DB -> LDAP synchronization


elecharny:
There is a compilation error

PR #376, opened 2025-03-05
Add a DB -> LDAP sync test

### LINKED ISSUE #369: Add a new test for DB -> LDAP syncrhonization
We have plenty of tests for Ldap -> (whatever) but no tests for DB ->LDAP synchronization.

elecharny:
Closes #369 A test was added to check that we can synchronize a DB with a LDAP destination. Both synchronous and clean mode are tested.


elecharny:
This PR is obsolete and will be replaced by a new one

PR #381, opened 2025-03-07
Fix for #369 Added a test for DB -> LDAP synchro

elecharny:


elecharny:
Clsed and replaced by PR #382

PR #397, opened 2025-04-25
Force compression alg for Debian package

coudot:
Back to .xz to support older Debian versions


guimard:
You may delete `override_dh_builddeb`: the default behavior is xz but can adapt its behavior depending on the source compression

coudot:
I added it because it was using zst, which is not compatible with Debian 11

I don't understand your comment, how fix compression do xz without override?

guimard:
The compression depends on target _(which should be the machine used to build package)_. In your case, you can use this workaround but this doesn't guarantee that package will really work on different Debian version

coudot:
Thanks for the answer. As the current Debian packaging is just used to install files in the filesystem, we use only one package for all Debian versions.

PR #412, opened 2025-07-28
Fix systemd units

EwenBara:


davidcoutadeur:
Already done in PR #443 and issue #413

Issue #430, opened 2025-09-23
In asynchronous mode with LDAP the entry get() should be blocking


elecharny:
When doing an asynchronous synchronization between two LDAP servers, the code uses a persistent search, then do get the result using this line of code:

```
searchResponse = sf.get(1, TimeUnit.NANOSECONDS);
```

There is no reason to get out of the get after only one nanosecond: We should leverage the Future and block until we get some data.

The current code exit immediately, and later on a 5 seconds wait is done before retrying to fetch entries, which delays the synchronization. The whole logic of PersistentSearch is really to react on a DIT change.

PR #432, opened 2025-09-24
Fix for issue #430

elecharny:
A one line change to make the PersistentSearch` working without a sleep, making any modifications done in the origin server immediately injected into the destination server.

The sleep is still there if for some reason the `getNextId()` call returns null.


elecharny:
Applied in the current branch for 2.3, in a different way.

PR #445, opened 2025-10-29
Bumped up the Maven plugin versions, bumped up the logback versio  to …

elecharny:
…solve a vulnerability (see https://github.com/lsc-project/lsc/security/dependabot/93)


elecharny:
Applied partially, and more updates have been made

PR #446, opened 2025-11-03
Configuration file fix

elecharny:
Added a fix for infering the logback.xml and other configuration files if we call LSC with a `-f <file>` parameter.


elecharny:
Replaced by a more accurate PR

PR #452, opened 2025-11-21
Added new methods getting a String parameter instead of a int

elecharny:
In order to avoid issues when using a String value for the UserAccessControl value, when it's expected to be an int, 3 methods have been added that allows a String top be used, which get converted to an int before calling the original method.


coudot:
It seems you are adding a new file instead of patching existing file

xavierba:
Tested with a previously failing configuration, it now works.

PR #462, opened 2026-01-06
Fix for PR 461

elecharny:
Added a method that takes the already cloned element to avoid cloning it again.


elecharny:
Closing as the duplicated as the `calculateModifications(Task task, IBean srcBean, IBean dstBean)` method is only called by some tests.

elecharny:
Already applied in another PR

Issue #465, opened 2026-01-19
Regression: when using the 'all' keyword for tasks execution, async and sync tasks are executed


elecharny:
In version before 2.2, it was not allowed to execute async tasks and clean/sync tasks at the same time.

It's still forbidden when specifying tasks by their name, but when using `-a all`, then it executes all the tasks, including the sync and clean ones.

This is a clear regression introduced in 2.2.

davidcoutadeur:
Indeed.

`-a` means asynchronous task (create + update)
`-s` means synchronous task (create + update)
`-c` means clean task (delete)

So a `-a` should never launch a sync or clean task

elecharny:
After a deeper analysis, it does not seem to be a regression. When calling LSC with `-a all`, we don't check the tasks status, so the check for sync/clean tasks is not leveraged, as the list of sync and clean tasks are empty. However, the `all` keyword get expanded later and without any check on each task's status.

elecharny:
I have two patches for that.

In `SimpleSynchronize.java`, while running `processTasks` method, we check if we are in async mode when we have a `all` keyword, and we only execute async tasks. 

```
	private int processTasks(List<String> tasks, Task.Mode mode) {
...
            for (TaskType taskType:LscConfiguration.getTasks()) {
		Task task = cache.get(taskType.getName());
				
		// Check the task type and discard it if it does not match with the requested mode
                if (task.getSourceService() instanceof IAsynchronousService
                        || task.getDestinationService() instanceof IAsynchronousService) {
                    // This is a n async task, only execute it in async mode, otherwise ignore it
                    if(mode == Task.Mode.async) {
                        launchTask |= processTask(task, mode);
                        nbLaunchedTasks++;
                    }
                } else {
                    // It's a sync or cleanup task only execute it in sync or cleanup mode
    	            if(mode != Task.Mode.async) {
                        launchTask |= processTask(task, mode);
			nbLaunchedTasks++;
                    }
                }
            }
...
```

It has to be combined with another patch in method `launch`:

```
		// Process the three different type of tasks, and get the number of executed tasks.
		// If this number is negative, it means there was a failure in the execution of
		// at least one task.
		int nbLaunchedSyncTasks = processTasks(syncTasks, Task.Mode.sync);
		int nbLaunchedCleanTasks = processTasks(cleanTasks, Task.Mode.clean);
		
		if (nbLaunchedSyncTasks + nbLaunchedCleanTasks > 0 )
		{
[...]

PR #466, opened 2026-01-19
Added some protection against the execution of async tasks when sync/…

elecharny:
Fix for issue #465


elecharny:
Already fixed in #500

PR #483, opened 2026-04-07
Configuration fix #467

### LINKED ISSUE #467: When using the -f option to specify the location of the configuration file, it's used as a directory base for the logging configuration file
When you use something like `lsc -f /blah/conf.xml`, the logging configuration file `logback.xml` is looked up into the `/blah/conf.xml/logback.xml ` location.

We should instead either look for this file in the parent directory (`/blah`) or use the default `etc` directory.

elecharny:
Fixes #467


davidcoutadeur:
PR has too much conflicts. The clean PR is in #499

Issue #460, opened 2026-01-06
Use an event-loop to process tasks


elecharny:
Currently we use a ThreadExecutor to process a task against a list of entries. If we have many entries, that will lead to the creation of as many `SynchronizeEntryRunner` instance which we be queued internally, consuming a lot of memory.

It would be much better to implement an event-loop system where the executor ask for more entry to process when some thread is available. We would then fetch the entry on demand, instead of storing it in memory.

elecharny:
I have something working with the following code, which still need some work. 
```
	public void run() {
		counter = new InfoCounter();

		SynchronizeThreadPoolExecutor threadPool = new SynchronizeThreadPoolExecutor(abstractSynchronize.getThreads());

		Entry<String, LscDatasets> nextId = null;
		boolean isInitDone = false;
		
		IAsynchronousService aService = null;
		boolean fromSource = true;
		
		if (task.getDestinationService() instanceof IAsynchronousService) {
			aService = (IAsynchronousService) task.getDestinationService();
			fromSource = false;
		} else if (task.getSourceService() instanceof IAsynchronousService) {
			aService = (IAsynchronousService) task.getSourceService();
		} else {
			LOGGER.error("LSC should never reach this point! Please consider debugging the code because "
			        + "we are trying to launch an asynchronous sync without any asynchronous service!"); 
			return;
		}

		LOGGER.debug("Asynchronous synchronize {}", task.getName());
		
		execute(aService, fromSource);
		/*
		boolean interrupted = false;
		
		while (!interrupted) {
			nextId = aService.getNextId();
			
			if (nextId != null) {
				threadPool.runTask(
				        new SynchronizeEntryRunner(task, counter, abstractSynchronize, nextId, fromSource));
			}
		}
		} catch (LscServiceException e) {
			counter.incrementCountError();
			abstractSynchronize.logActionError(null, nextId, e);
		}
		*/

		try {
			threadPool.shutdown();
			threadPool.awaitTermination(abstractSynchronize.getTimeLimit(), TimeUnit.SECONDS);
		} catch (InterruptedException e) {
			LOGGER.error("Tasks terminated according to time limit: " + e.toString(), e);
			LOGGER.info("If you want to avoid this message, " + "increase the time limit by using dedicated parameter.");
		}
	}
	
	private void execute(IAsynchronousService asyncService, boolean fromSource) {
	    int nThreads = abstractSynchronize.getThreads();
[...]

PR #487, opened 2026-04-08
o Various code refactoring

elecharny:
Fix for issues #465 and #460:

o Handle the deletion of entries when a async task is ran on a LDAP server
o Replaced  the poll loop in async mode by an event driven call 
o Replaced some Cursor code to use the 'for' new syntax o Simplified the declaration of new array
o Some code refactoring for clarity
o Added some time counter and log for the refresh phase 
o Prepared a handling for the end of the refresh phase in case we want to do a cleanup
o Replaced a declaration of a pattern by a static declaration


etc/lsc-clean-orange-ucs-pds.xml
coudot, line 1: Is this file needed? Or just an example? In any case, it should be renamed. And if it is an example, it could go in sample/

elecharny: No, it will be removed in a commit to come.

elecharny:
To rewrite

PR #488, opened 2026-04-08
Handle the '*' attribute in fetched attributes

### LINKED ISSUE #177: Possibility to specifiy * in fetchedAttributes to synchronise all attributes of entries
For now we must define all attributes in fetchedAttributes, in source and destination.

An evolution would be to allow * value in fetchedAttributes to get all attributes of the entry and sync them.

elecharny:
Fix for NGI #177 : handle the '*' attribute in fetched attributes


davidcoutadeur:
This PR has too many conflicts, and should be moved to a new, cleaned PR.

Issue #489, opened 2026-04-10
Make sure LSC retries in async mode when the connection is shutdown


elecharny:
When the connection is shutdown for any reason, LSC just fails and has to be relaunched 'by hand'. It would be convenient to have an infinite retry to workaround disconnections.

elecharny:
We should add a number of retry parameter `cnxRetryNumber` to the connection configuration, with -1 for infinite number. The default will be 3 (which is the actual default), and a 'cnxRetryDelay' which is defaulting to 30s at the moment. 

It's for async services only, and LDAP only. Also verify it works for source and destination

Check that it works with unit tests.

elecharny:
The associated PR should be reworked and renamed.

Issue #486, opened 2026-04-08
Add a way to delete entries in destination in LDAP async mode


elecharny:
When running in async mode, on a LDAP source server, we would like the deleted entries to be also deleted in the destination.
Currently, those deletion are ignored, although we know about the DN being deleted.
The idea is to enforce the deletion of such entries in the destination based on the received DN.

davidcoutadeur:
I don't see any flag for enabling this feature.
If I understand correctly, the feature would now delete entries in destination as soon as they are deleted in the source.
This is a huge change of paradigm. The users currently using the async connector are expecting entries **not** to be deleted. Thus I think we should provide this feature as an option, disabled by default.

davidcoutadeur:
Also, the current PR seems to work only for

async LDAP -> LDAP

It would be more generic (and maybe not so complex) to make it work for : 

async LDAP -> any

davidcoutadeur:
After discussion, it appears that the feature can't be implemented in a generic way:

We need the source pivot ids to determine the destination mainIdentifier, but we don't have these pivots ids as they are not present in the syncrepl delete operation, nor in the source entry as it has already been deleted.

So this feature will be closed and not included in lsc-core. The PR and the associated branch can be preserved for people willing to implement a specific patch on their instance.

PR #490, opened 2026-04-10
Retry cnx

elecharny:
Fix for issues #489 and #486


elecharny:
Replaced by 489-retry

PR #491, opened 2026-04-10
Improvments

elecharny:
Many minor fixes and improvements:

* Potential NPE fix
* Test to show that clean+sync work together
* Added Conditions handling in annotations
* Added some debug messages on deletion
* Added a function that use the read entry instead of fetching it from the source server


elecharny:
To rewrite rebased on master

PR #494, opened 2026-04-24
Patch CVE-2024-52046

jusabatier:
I use this project to build a docker image that is scanned by trivy, and it signal a critical CVE :

Severity | CVE | Package | Installed | Fixed | Target
-- | -- | -- | -- | -- | --
CRITICAL | CVE-2024-52046 | org.apache.mina:mina-core | 2.2.3 | 2.2.4, 2.1.10, 2.0.27 | Java

This PR solve this CVE by force use of mina-core 2.2.5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -756,6 +756,16 @@
 		</pluginRepository>
 	</pluginRepositories>
 
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.apache.mina</groupId>
+				<artifactId>mina-core</artifactId>
+				<version>2.2.5</version>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
+
 	<dependencies>
 		<dependency>
 			<groupId>org.forgerock.ce.opendj</groupId>


elecharny:
Hi,

FTR, LSC is not impacted by this CVE.

We will upgrade nevertheless.

elecharny:
Also note that we have just released MINA 2.2.6 that fixes two more CVE in the exact same feature.

elecharny:
Ok, so reality is that it's a useless upgrade. The MINA dependency relates to the Apache LDAP API used in the Apache Directory Server test platform.

We will upgrade both components when they will have been released, hopefully in the coming weeks.

But again: they aren't impacted by the CVEs.

jusabatier:
Thanks for the clarification, and sorry if my message sounded alarmist.

I was mainly investigating a scanner result on a bundled dependency and trying to understand whether it was actually relevant in the real runtime context.

Thanks again for the quick feedback and for the planned future update.

elecharny:
Not a problem! I'd rather close a PR that is not necessary than not receiving a PR addressing a real problem ;-)

This is what is 'triage' is about.

FTR, the MINA CVE you mentionned was nearly 2 years old, I'm one of the MINA maintainer, so I have a pretty good grap on the issue. However, a CVE is always a turn off for users, so better get rid of them even if they don't have any impact on the project.

You can't imagine the number of people being frightened by a project with a CVE in a dependency :-) Rightful so, I would say! 

Many thanks for the report btw.

PR #501, opened 2026-05-05
Fix for star in attributes

### LINKED ISSUE #177: Possibility to specifiy * in fetchedAttributes to synchronise all attributes of entries

elecharny:
Fixes #177

### DIFF
--- a/src/main/java/org/lsc/jndi/AbstractSimpleJndiService.java
+++ b/src/main/java/org/lsc/jndi/AbstractSimpleJndiService.java
@@ -47,6 +47,7 @@ package org.lsc.jndi;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -61,6 +62,7 @@ import javax.naming.directory.Attribute;
 import javax.naming.directory.SearchControls;
 import javax.naming.directory.SearchResult;
 
+import org.apache.directory.api.util.Strings;
 import org.lsc.Configuration;
 import org.lsc.LscDatasets;
 import org.lsc.beans.IBean;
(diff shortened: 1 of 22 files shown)


src/main/java/org/lsc/jndi/AbstractSimpleJndiService.java
```diff
@@ -163,24 +165,76 @@ public AbstractSimpleJndiService(final Properties serviceProps) throws LscServic
...
 		baseDn = ldapService.getBaseDn();
-		filterIdSync = (ldapService.getOneFilter() != null ? ldapService.getOneFilter().trim() : ldapService.getGetOneFilter().trim());
-		filterAll = (ldapService.getAllFilter() != null ? ldapService.getAllFilter().trim() : ldapService.getGetAllFilter().trim());
+
+		// Get the allFilter, and trim it
+		filterAll = (ldapService.getAllFilter() != null ?
+		    ldapService.getAllFilter().trim() :
+		    ldapService.getGetAllFilter().trim());
+
+		// Deal with the case were we don't have a oneFilter, we will default to allFilter
```
davidcoutadeur: I really don't see the point of defaulting oneFilter to allFilter.

Indeed, there is no way LSC could guess the oneFilter correctly.

Could you rewrite this function or explain the use case of defaulting oneFilter to allFilter?

src/main/resources/schemas/lsc-core-2.2.xsd
```diff
@@ -326,8 +326,10 @@
 						<xsd:element name="allFilter" type="xsd:string" />
 					</xsd:choice>
 					<xsd:choice>
-						<xsd:element name="getOneFilter" type="xsd:string" />
-						<xsd:element name="oneFilter" type="xsd:string" />
+						<xsd:element name="getOneFilter" type="xsd:string"
+							minOccurs="0"/>
```
davidcoutadeur: Same here: I think getOneFilter / oneFilter should be mandatory (no minOccurs="0")
Could you explain the use case or remove the minOccurs?

davidcoutadeur (changes requested):
I have reviewed this PR, could you do the changes @elecharny ?

src/main/java/org/lsc/jndi/AbstractSimpleJndiService.java
elecharny: It's an anticipation for the coming changes: being able to fetch all the entries at once, to avoid doing N requests.
That was a change I made when doing some performance tests for Orange, and it speed up the LSC execution by nearly x2.

davidcoutadeur: Change has been reverted and will be reinjected in its own PR

src/main/resources/schemas/lsc-core-2.2.xsd
davidcoutadeur: Change has been reverted and will be reinjected in its own PR

PR #503, opened 2026-05-07
Add LdapRestDstService: write-through ldap-rest HTTP API

guimard:
## Summary

Adds a new `IWritableService` implementation, `LdapRestDstService`,
that lets LSC push modifications through a JSON/HTTP API instead of
binding directly to LDAP. The target API is
[ldap-rest](https://github.com/linagora/ldap-rest), which exposes an
LDAP-compatible REST surface on top of an LDAP back-end, plus
fine-grained ACLs, schema validation, audit logging, and webhooks.

## Motivation

Several deployments need an extra layer between LSC and the directory:

- **ACLs at attribute granularity** that the LDAP back-end alone
  cannot express;
- **Audit trail** tied to a service identity (HMAC-SHA256 or bearer);
- **Hooks downstream** (badge / mail / billing system) on each
  successful change;
- **Schema validation** beyond LDAP's `objectClass` constraints.

In production, this is currently shipped as an external plugin via
`<pluginDestinationService implementationClass=…>` (see
[linagora/ldap-rest#76](https://github.com/linagora/ldap-rest/pull/76)).
Folding it into LSC core lets the same configuration live as a typed
`<ldapRestDestinationService>` element, parsed via JAXB, with proper
schema validation in `lsc.xml` rather than a `<xsd:any>` opaque
block.

## What this PR adds

**Schema (`src/main/resources/schemas/lsc-core-2.2.xsd`):**
- new `ldapRestAuthType` (choice of `<bearer>` or
  `<hmacServiceId>`+`<hmacSecret>`);
- new `ldapRestDestinationServiceType` extending `serviceType`
  with `baseUrl`, `resourceType`, `apiPrefix`, `auth`, `timeoutMs`,
  `retries`, `writeDatasetIds`;
- adds `<ldapRestDestinationService>` to the destination-service
  choice in `taskType` (alongside `<ldapDestinationService>`,
  `<pluginDestinationService>`, …).

JAXB types are generated automatically via the existing
`maven-jaxb2-plugin` configuration — no plugin tweaks needed.

**Code (`src/main/java/org/lsc/service/ldaprest/`):**
- `LdapRestDstService` — `IWritableService`, parses
  `task.getLdapRestDestinationService()` (typed JAXB) and dispatches
  CREATE/UPDATE/DELETE/MODRDN to ldap-rest endpoints;
- `LdapRestClient` — HTTP wrapper with retry on 5xx/IOException
  and capped exponential backoff;
- `LdapRestAuth`, `BearerAuth`, `HmacAuth` — pluggable auth
  strategies; `HmacAuth` is wire-compatible with ldap-rest's
  server-side verification (`METHOD|PATH|TS|sha256(body)`);
- `ResourceMapper` — endpoint routing per resource family
  (`groups`, `organizations`, flat) with RFC 4514 escape handling;
- `ModificationTranslator` — `LscModifications` → JSON converter
  with deterministic insertion-order output (signature stability).

**Wiring (`org.lsc.configuration.LscConfiguration`):**
- `getDestinationService`, `setDestinationService`,
  `getServiceImplementation` updated to recognise the new type and
  map it to `LdapRestDstService.class`.

**Tests (`src/test/java/org/lsc/service/ldaprest/`):**
- 91 JUnit 5 tests, covering auth signing (cross-impl HMAC vector),
  HTTP retries, JSON shape stability, RFC 4514 DN parsing, full
  CREATE/UPDATE/DELETE/MODRDN flows for users / groups /
  organizations, group member endpoint routing including
  attribute-wide DELETE reconciliation. WireMock 3.9.1 is added
  as a `<scope>test</scope>` dependency.

**Documentation (`doc/ldap-rest-destination-service.md`):**
- service overview, configuration reference, sample task, supported
  operations, identifier mapping, known limitations.

## Files added / modified

```
A doc/ldap-rest-destination-service.md
M pom.xml                                                    (+wiremock test dep)
M src/main/java/org/lsc/configuration/LscConfiguration.java  (3 hooks)
M src/main/resources/schemas/lsc-core-2.2.xsd                (2 types + 1 choice element)
A src/main/java/org/lsc/service/ldaprest/LdapRestDstService.java
A src/main/java/org/lsc/service/ldaprest/LdapRestClient.java
A src/main/java/org/lsc/service/ldaprest/LdapRestAuth.java
A src/main/java/org/lsc/service/ldaprest/BearerAuth.java
A src/main/java/org/lsc/service/ldaprest/HmacAuth.java
A src/main/java/org/lsc/service/ldaprest/ResourceMapper.java
A src/main/java/org/lsc/service/ldaprest/ModificationTranslator.java
A src/test/java/org/lsc/service/ldaprest/LdapRestDstServiceConfigTest.java
A src/test/java/org/lsc/service/ldaprest/LdapRestDstServiceIntegrationTest.java
A src/test/java/org/lsc/service/ldaprest/LdapRestClientTest.java
A src/test/java/org/lsc/service/ldaprest/LdapRestAuthTest.java
A src/test/java/org/lsc/service/ldaprest/ResourceMapperTest.java
A src/test/java/org/lsc/service/ldaprest/ModificationTranslatorTest.java
```

## Backward compatibility

Existing `<ldapDestinationService>`, `<pluginDestinationService>`,
`<databaseDestinationService>`, `<multiDestinationService>` and
`<xaFileDestinationService>` configurations parse and run
unchanged — only an additional choice was added next to them.
None of the existing classes were modified beyond the three
explicit dispatch hooks in `LscConfiguration`.

## Known limitations

- Binary attributes are not supported (values that fail strict
  UTF-8 decoding raise `LscServiceException`); UTF-8 byte arrays
  are silently accepted.
- Each LSC modification produces one HTTP call; no batching.
  Group `member` reconciliation fans out one DELETE per current
  member.
- `REPLACE` on the group `member` attribute is best-effort
  (add-all-new without computing delete-missing). Use explicit
  ADD/DELETE for deterministic group sync.
- This service is **destination-only**: `getBean` /
  `getListPivots` return empty structures.

## Test plan

- [x] `mvn -B test` passes on JDK 17 (532 tests, 91 new, 0
      failures, 0 errors, 2 pre-existing skipped)
- [x] new tests cover all four LSC operations across `users`,
      `groups`, `organizations`
- [x] HMAC vector cross-checked against the ldap-rest Node server
      (same fixed inputs → same hex signature)
- [x] schema generates JAXB types under
      `target/generated-sources/xjc/org/lsc/configuration/`
- [x] no existing test was modified
- [x] no existing class was modified except the three explicit
      dispatch hooks in `LscConfiguration`

### DIFF
--- a/doc/ldap-rest-destination-service.md
+++ b/doc/ldap-rest-destination-service.md
@@ -1,0 +1,153 @@
+# LDAP REST destination service
+
+`LdapRestDstService` is an `IWritableService` implementation that
+sends modifications through an HTTP/JSON API instead of writing
+directly to an LDAP directory. It targets the
+[ldap-rest](https://github.com/linagora/ldap-rest) project, which
+exposes an LDAP-compatible REST API on top of an LDAP back-end and
+adds fine-grained ACLs, schema validation, audit logging and
+downstream webhooks.
+
+## When to use it
+
+Use `<ldapRestDestinationService>` when you want LSC to feed
+identity changes into a system that:
+
+- enforces ACLs on a per-resource / per-attribute basis;
+- runs business logic (provisioning, e-mail, badge…) on each
+  change via webhooks;
+- needs a strict audit trail tied to a service identity;
+- speaks JSON rather than LDAP for operational reasons.
+
+When the destination is a plain LDAP directory the existing
+`<ldapDestinationService>` is more efficient (single TCP
+connection, no JSON round-trip).
+
+## Configuration
+
+Drop a `<ldapRestDestinationService>` element in your task. Fields:
+
+| element             | required | default  | meaning                                                                  |
+|---------------------|----------|----------|--------------------------------------------------------------------------|
+| `<name>`            | yes      | —        | task-unique name, inherited from `serviceType`                           |
+| `<connection>`      | yes      | —        | inherited from `serviceType`; **ignored** at runtime by this service     |
+| `<baseUrl>`         | yes      | —        | scheme+host+port of the ldap-rest server, e.g. `https://ldap-rest.example.org` |
+| `<resourceType>`    | yes      | —        | one of `users`, `groups`, `organizations`, or any flat resource your ldap-rest server exposes |
+| `<apiPrefix>`       | no       | `/api`   | path prefix of the API (e.g. `/api/v2`); use `/` for no prefix           |
+| `<auth>`            | yes      | —        | authentication block, see below                                          |
+| `<timeoutMs>`       | no       | `10000`  | per-request timeout in milliseconds                                      |
+| `<retries>`         | no       | `3`      | number of retries on HTTP 5xx or I/O errors                              |
+| `<writeDatasetIds>` | no       | empty    | dataset ids this service is allowed to write (`<string>` children)       |
+
+The schema requires `<connection>` because every service inherits
+it from `serviceType`. `LdapRestDstService` does not actually use
+the bound connection; operators usually point it to a stub LDAP
+connection or any existing one to keep the schema valid.
+
+### Authentication
+
+Two auth strategies are supported. Pick one inside `<auth>`:
+
+```xml
+<auth>
+  <bearer>${LDAP_REST_TOKEN}</bearer>
+</auth>
+```
+
+or
+
+```xml
+<auth>
+  <hmacServiceId>lsc-prod</hmacServiceId>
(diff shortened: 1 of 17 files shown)


elecharny:
This would be way better as a plugin living in its own repo, like the Fusion Directory plugin (https://github.com/lsc-project/lsc-fusiondirectory-plugin/tree/master).

It's very specific, and does not really deserve to be included in LSC core, with a configuration modification. This is the reason we have plugins.

PR #515, opened 2026-06-24
allow javascript in LDAP filters (#79)

### LINKED ISSUE #79: allow javascript in LDAP filters
It would be nice to have the possibility to use javascript in LDAP filter, even if srcBean and dstBean are not available at this place.
For example for this use case:

```
var today = new Date();
// compute todayGeneralizedTime
var result = "(&objectClass=inetOrgPerson)(personLeaveDate >= " + todayGeneralizedTime + "))";
result
```

davidcoutadeur:
Fixes #79

### DIFF
--- a/etc/lsc.xml-sample
+++ b/etc/lsc.xml-sample
@@ -1,5 +1,5 @@
 <?xml version="1.0" ?>
-<lsc xmlns="http://lsc-project.org/XSD/lsc-core-2.2.xsd" revision="0">
+<lsc xmlns="http://lsc-project.org/XSD/lsc-core-2.3.xsd" revision="0">
 <!--
      See full documentation online:
      https://lsc-project.org/documentation/latest/index_configuration.html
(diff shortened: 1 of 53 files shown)


davidcoutadeur:
TODO:

- [x] test if the engine works with GraalJS
- [x] find a way to pass the current pivotIDs to the JS engine in a variable
- [x] add the possibility to use an external script in the filter
- [x] find a way not to escape the existing filters: done with solution: adding new tags: allEntriesFilter, oneEntryFilter, and cleanEntryFilter
- [x] write the doc for using scripts in filters (especially pivotsIDs, usable interfaces,...) + upgrade notes
- [x] treat the cleanEntryFilter case
- [x] add tests to deal with the cleanEntryFilter case
- [x] test if the script filter works with rjs + GraalJS
- [x] mark old options as deprecated

davidcoutadeur:
I have finished the review. The PR will be merged as soon as I complete the TODO list above. (doc + more tests)

PR #517, opened 2026-06-26
Retry indefinitively when we get a connection error in async mode. Fi…

elecharny:
…xed a potential NPE. Made the oneFilter optional in annotations


elecharny:
Replaced by another PR (489-retry)

PR #526, opened 2026-07-08
Remove deprecated finalize

### LINKED ISSUE #475: Remove the finalize() methods from the code
The `finalize()` method has been deprecated in Java 8 and will be removed at some point.

It's implemented in the following classes:

- `AbstractSimpleJndiService`
- `JndiServices`

We should remove those calls, and explicitly call a cleanup method instead

Kcchouette:
Fix #475 

- Implement AutoCloseable on JndiServices
- Replace finalize() with proper close() method
- Simplify AbstractSimpleJndiService.close() delegation
- Add unit tests for close() behavior


elecharny:
Patch applied.
Thanks!

Issue #341, opened 2025-02-18
A clean phase may eat a lot of memory


elecharny:
While doing a clean (ie deleting all the entries present in the target but not anymore in the source), we end doing that:

* Fetch all the IDS from the source 
* for each ID
  * create a Runnable
  * create a Thread with this Runnable
  * put it into the ThreadPool for execution

If we have thousands of ID to delete, that will create as many Threads instance for nothing. Assuming we are using an Executor that will create a new Thread when a slot is available for its execution, we can simply pass the Runnable.

If the Thread is created in order to give it a name, then it's better to use a ThreadFactory that will create the Thread and give it a name when the executor is ready.

Kcchouette:
For this issue, removing this line should be sufficient?

https://github.com/lsc-project/lsc/blob/eacf33d2e58d3be36e2cde93e917402b81db5db4/src/main/java/org/lsc/SynchronizeThreadPoolExecutor.java#L52

elecharny:
Actually, a better way to deal with this issue is to use an event loop that fetches the next entry when a thread  in the thread pool is available. This is what we have implemented in the code.
This way we don't stack thousands of Thread instances waiting for a thread to be available, we create then only when needed.

Here is the code:

```

        // Supplier of tasks, will be called by the thread pool,
        // returns null if no more tasks to run
        Supplier<Callable<Object>> supplier = () -> {
            try {
                Entry<String, LscDatasets> nextId;

                // Fetch an entry from the source
                lock.lock();
                
                try {
                    // This is a blocking call, we may wait until we get
                    // a response
                    nextId = asyncService.getNextId();
                } finally {
                    lock.unlock();
                }
                
                if (nextId == null) {
                    // we're done
                    return null;
                }
                
                // Now compute the modifications
                return new Callable<Object>() {
                    @Override
                    public String call() throws Exception {
                        // Process the entry
                        SynchronizeEntryRunner entryRunner = new SynchronizeEntryRunner(
                                task, counter, abstractSynchronize, nextId, fromSource);
                        
                        entryRunner.run();
                        
                        return nextId.getKey() + " processed";
                    }
                };
            } catch (LscServiceException e) {
                counter.incrementCountError();
                abstractSynchronize.logActionError(null, null, e);
            }
            
            return null;
        };

        // Thread pool, each thread loops until the task supplier returns a null Callable
        // Basically, fetch an entry from the source, and process the associated callable
        // It's a fixed sized thread pool, once created the threads are up forever
[...]

PR #530, opened 2026-07-15
Applied the suggested modification by kcchouette for issue #341

elecharny:


src/main/java/org/lsc/SynchronizeThreadPoolExecutor.java
```diff
@@ -19,14 +19,18 @@
...
 	BlockingQueue<Runnable> queue;
 
 	/** Default logger */
 	final Logger LOGGER = LoggerFactory.getLogger(SynchronizeThreadPoolExecutor.class);
 
 	public SynchronizeThreadPoolExecutor(int threads) {
+	    this(threads, Math.max(100, threads * 10));
+	}
+	
+    public SynchronizeThreadPoolExecutor(int threads, int maxQueue) {
```
Kcchouette, line 33: `maxQueue` -> `queueCapacity`

Kcchouette: mmh, it seems still a draft pr

elecharny: I'm having difficulties with Java 17. Will fix the maxQueue name tomorrow

elecharny: Hi @Kcchouette, here is what I'm coming with:

```
/**
 * This object is managing LSC tasks. It is now a wrapper for JDK 
 * ThreadPoolExecutor but may rely on a different implementation
 * 
 * The queue size is computed based on the number of requested threads, 
 * and have a minimum value of 100.
 * 
 * The RejectedExecution handler will simply block when the queue is full,
 * so the caller will wait until a thread completes and consume the queue.
 * @author Sebastien Bahloul &lt;seb@lsc-project.org&gt;
 *
 */
public class SynchronizeThreadPoolExecutor extends ThreadPoolExecutor {

	static long keepAliveTime = 60;
	
	/** The default queue size. We default to a small size */
	static final int DEFAULT_QUEUE_CAPACITY = 100;

	BlockingQueue<Runnable> queue;

	/** Default logger */
	final Logger LOGGER = LoggerFactory.getLogger(SynchronizeThreadPoolExecutor.class);

	public SynchronizeThreadPoolExecutor(int threads) {
	    // The queue size is computed based on the number of threads
	    // A first approach is to require 10 slots in the queue per requested thread
	    this(threads, Math.max(DEFAULT_QUEUE_CAPACITY, threads * 10));
	}
	
    public SynchronizeThreadPoolExecutor(int threads, int queueCapacity) {
		super(threads, threads, keepAliveTime, TimeUnit.SECONDS, new ArrayBlockingQueue<Runnable>(queueCapacity), 
		        new RejectedExecutionHandler() {
        			public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
        				// this will block if the queue is full
        				try {
        						executor.getQueue().put(r);
        					} catch (InterruptedException e) {
        						Thread.currentThread().interrupt();
        						throw new RuntimeException(e);
        					}
        				}
        			}
        		);
		queue = getQueue(); 
	}
```

PR #532, opened 2026-07-16
Option to bypass "getOne" step (#163)

### LINKED ISSUE #163: Option to bypass "getOne" step
For small directories, or when few attributes are synced, we could fetch all values in the "getAll" step and bypass the "getOne" step for each entry in source.

davidcoutadeur:
Fixes #163

### DIFF
--- a/src/main/java/org/lsc/jndi/SimpleJndiSrcService.java
+++ b/src/main/java/org/lsc/jndi/SimpleJndiSrcService.java
@@ -54,6 +54,7 @@ import javax.naming.directory.SearchControls;
 
 import org.lsc.LscDatasets;
 import org.lsc.beans.IBean;
+import org.lsc.beans.SimpleBean;
 import org.lsc.configuration.LdapSourceServiceType;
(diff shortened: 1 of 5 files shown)


src/main/java/org/lsc/jndi/SimpleJndiSrcService.java
```diff
@@ -182,7 +188,17 @@ public IBean getBean(Task task, final String pivotName, final LscDatasets pivotA
 				}
 			}
 
-			return this.getBeanFromSR(get(task, pivotName, pivotAttributes, searchString, scriptAsFilter), srcBean);
+			if( isBypassOneEntry() )
```
Kcchouette: better to use the var `bypassOneEntry`?

davidcoutadeur: Indeed, I fix the PR

elecharny (approved):
Sounds good to me

PR #535, opened 2026-07-17
Fixes #486: delete entries in async mode

### LINKED ISSUE #486: Add a way to delete entries in destination in LDAP async mode

elecharny:
Fixes #486

### DIFF
--- a/src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
+++ b/src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
@@ -3,7 +3,6 @@ package org.lsc.runnable;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
 
 import org.lsc.AbstractSynchronize;
 import org.lsc.beans.InfoCounter;
@@ -14,11 +13,12 @@ import org.lsc.Task;
 import org.lsc.beans.BeanComparator;
 import org.lsc.beans.IBean;
 import org.lsc.exception.LscServiceCommunicationException;
+import org.lsc.jndi.SimpleJndiDstService;
+import org.lsc.service.SyncReplSourceService;
 import org.lsc.utils.ScriptingEvaluator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.lsc.Hooks;
-import org.lsc.beans.syncoptions.ISyncOptions.OutputFormat;
 
 /**
  * @author sbahloul
@@ -41,8 +41,32 @@ public class SynchronizeEntryRunner extends AbstractEntryRunner {
 	@Override
 	public void run() {
 		counter.incrementCountAll();
-		try {
-			run(abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(), id.getKey(), id.getValue(), true, fromSource));
+
+        try {
+            // Deal with the special case of deleted entries, when the destination service
+            // is a LDAP server. We *must* base the deletion on the source entry's DN
+            // because we have nothing else to identify the entry (the pivot is not present).
+            if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
+                    && (task.getDestinationService() instanceof SimpleJndiDstService)) {
+                // Create a modification for the Delete operation
+                LscModifications lm = new LscModifications(LscModificationType.DELETE_OBJECT, task.getName());
+
+                // Get the source entry DN which is stored in the entry's key
+                String dn = id.getKey();
+
+                // Transform it to get rid of the base, as we are using JNDI which uses a context
+                String newDn = ((SimpleJndiDstService)task.getDestinationService()).getJndiServices().
+                    rewriteBase(dn);
+
+                lm.setMainIdentifer(newDn);
+
+                // And now, delete the entry
+                task.getDestinationService().apply(lm);
+            } else {
+                IBean data = abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(),
+                        id.getKey(), id.getValue(), true, fromSource);
+                run(data);
+            }
 		} catch (RuntimeException e) {
 			counter.incrementCountError();
 			abstractSynchronize.logActionError(null, id.getValue(), e);
--- a/src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
+++ b/src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
@@ -48,20 +48,7 @@ public class SynchronizeEntryRunner extends AbstractEntryRunner {
             // because we have nothing else to identify the entry (the pivot is not present).
             if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
                     && (task.getDestinationService() instanceof SimpleJndiDstService)) {
-                // Create a modification for the Delete operation
-                LscModifications lm = new LscModifications(LscModificationType.DELETE_OBJECT, task.getName());
-
-                // Get the source entry DN which is stored in the entry's key
-                String dn = id.getKey();
-
-                // Transform it to get rid of the base, as we are using JNDI which uses a context
-                String newDn = ((SimpleJndiDstService)task.getDestinationService()).getJndiServices().
-                    rewriteBase(dn);
-
-                lm.setMainIdentifer(newDn);
-
-                // And now, delete the entry
-                task.getDestinationService().apply(lm);
+               deleteEntry();
             } else {
                 IBean data = abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(),
                         id.getKey(), id.getValue(), true, fromSource);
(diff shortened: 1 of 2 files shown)


src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -41,8 +41,32 @@ public SynchronizeEntryRunner(final Task task, InfoCounter counter,
...
 	public void run() {
 		counter.incrementCountAll();
-		try {
-			run(abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(), id.getKey(), id.getValue(), true, fromSource));
+
+        try {
+            // Deal with the special case of deleted entries, when the destination service
+            // is a LDAP server. We *must* base the deletion on the source entry's DN
+            // because we have nothing else to identify the entry (the pivot is not present).
+            if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
```
davidcoutadeur: Not sure this is the best place to apply the deletion. IMO, it should be inside the synchronization task, at a lower place.

src/main/java/org/lsc/service/SyncReplSourceService.java
```diff
@@ -237,101 +245,176 @@ public IBean getBean(Task task, final String id, final LscDatasets pivotAttrs, b
...
+			searchRequest.addControl(searchContinuationControl);
+
+			// We will use the configured base DN and filter
+			searchRequest.setBase(new Dn(getBaseDn()));
+			searchRequest.setFilter(getFilterAll());
+			searchRequest.setDerefAliases(getAlias(ldapConn.getDerefAliases()));
+			searchRequest.setScope(SearchScope.SUBTREE);
+
+			// Reclaim all attributes so that we don't have to do it later.
+			//searchRequest.addAttributes(SchemaConstants.ALL_ATTRIBUTES_ARRAY);
```
davidcoutadeur: Why don't you set the admin defined attributes here (getAttrsId)?

Also, delete the comment if it is dead code.

src/main/java/org/lsc/service/SyncReplSourceService.java
```diff
@@ -237,101 +245,176 @@ public IBean getBean(Task task, final String id, final LscDatasets pivotAttrs, b
...
+				}
+			} else if ( searchResponse instanceof SyncInfoValue ) {
+				// This is the end of the refresh phase, log the info
+				SyncInfoValue syncInfoValue = (SyncInfoValue)searchResponse;
+				LOGGER.info("Refresh phase done: {} in {} ms", syncInfoValue.isRefreshDone(),
+				            System.currentTimeMillis() - refreshStart);
+
+				// Now we are done with the refresh, we need to cleanup the destination
+				// by removing the deleted entries
+				cleanupDestination();
```
davidcoutadeur: What's this for?

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -41,8 +41,32 @@ public SynchronizeEntryRunner(final Task task, InfoCounter counter,
...
+                String dn = id.getKey();
+
+                // Transform it to get rid of the base, as we are using JNDI which uses a context
+                String newDn = ((SimpleJndiDstService)task.getDestinationService()).getJndiServices().
+                    rewriteBase(dn);
+
+                lm.setMainIdentifer(newDn);
+
+                // And now, delete the entry
+                task.getDestinationService().apply(lm);
```
davidcoutadeur: The deletion seems enabled by default.
As this is a complete behaviour change (ie deletion of entries is guaranteed to be impossible for sync and async tasks), I propose to make this an option instead. For example: `allowAsyncEntryDeletion`. This option would be disabled by default

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -41,8 +41,32 @@ public SynchronizeEntryRunner(final Task task, InfoCounter counter,
...
+            if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
+                    && (task.getDestinationService() instanceof SimpleJndiDstService)) {
+                // Create a modification for the Delete operation
+                LscModifications lm = new LscModifications(LscModificationType.DELETE_OBJECT, task.getName());
+
+                // Get the source entry DN which is stored in the entry's key
+                String dn = id.getKey();
+
+                // Transform it to get rid of the base, as we are using JNDI which uses a context
+                String newDn = ((SimpleJndiDstService)task.getDestinationService()).getJndiServices().
```
davidcoutadeur: What if the destination is a database?
Does this work with an async_ldap -> database use case? If feasible, I think we should make it work with any destination.

davidcoutadeur (changes requested):
I have 2 major remarks about making this feature an option, and making it work with any destination.
I have also some other comments.

While reviewing, I have fixed the whole indentation, so take care to get back to the current revision in the branch.

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
elecharny: The reason it's done in the `run()` method and not in the `run(IBean)` method is that we won't fetch the bean at all: we have to delete it anyway.
Actually, we could create a dedicated method `delete()`, for clarity, but that won't change a thing:

```
        try {
            // Deal with the special case of deleted entries, when the destination service
            // is a LDAP server. We *must* base the deletion on the source entry's DN
            // because we have nothing else to identify the entry (the pivot is not present).
            if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
                    && (task.getDestinationService() instanceof SimpleJndiDstService)) {
                delete();
            } else {
                IBean data = abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(),
                        id.getKey(), id.getValue(), true, fromSource);
                run(data);
            }
...

	private void delete() throws LscServiceException {
        // Create a modification for the Delete operation
        LscModifications lm = new LscModifications(LscModificationType.DELETE_OBJECT, task.getName());

        // Get the source entry DN which is stored in the entry's key
        String dn = id.getKey();

        // Transform it to get rid of the base, as we are using JNDI which uses a context
        String newDn = ((SimpleJndiDstService)task.getDestinationService()).getJndiServices().
            rewriteBase(dn);

        lm.setMainIdentifer(newDn);

        // And now, delete the entry
        task.getDestinationService().apply(lm);
	}
```

src/main/java/org/lsc/service/SyncReplSourceService.java
elecharny: This is not exactly dead code ;-) But, yes, it has nothing to do there. Actually, it was part of a attempt to avoid fetching the attributes beforehand (a `getAll` without `getOne` for async mode). I left it commented to remember this is a future optimization...

So basically attributes will be fetched elsewhere.

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
elecharny: Very valid point. The problem is that we simply can't build the ID of the element to delete in the destination when the source deleted entry's DN does not contain  enough information to create this ID. And this is the only information we get when LDAP tells us one entry has been deleted: we get the DN, no attributes.

We then can't use the pivot to find the destination entry in all cases.

This is a serious limitation, but until LDAP provides all the attributes (not going to happen soon, I'm afraid), we are good to update the documentation to tell the users it's a LDAP to LDAP only feature.

src/main/java/org/lsc/service/SyncReplSourceService.java
elecharny: Nothing. Most certainly an early piece of code that has never been removed.

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
elecharny: I'd rather go in the other direction: add a flag to forbid deletion of entries.

RFC 4533 (syncrepl) and Persistent search control draft have always included the delete event, and the fact that we don't handle it is a missing feature. People should expect deletion to occur.

The LSC documentation actually says nothing about the fact that deletion never occur, which is a mistake (https://lsc-project.org/documentation/latest/service_source_asyncldap.html)

src/main/java/org/lsc/service/SyncReplSourceService.java
elecharny: Comment removed.

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
elecharny: A `delete()` method has been created

src/main/java/org/lsc/service/SyncReplSourceService.java
elecharny: My bad, this is for a cleanup phase that has to be implemented: we have processed added/modified/moved entries, we know nothing about entries that were deleted...

src/main/java/org/lsc/runnable/SynchronizeEntryRunner.java
```diff
@@ -43,18 +43,10 @@ public SynchronizeEntryRunner(final Task task, InfoCounter counter,
...
-            // because we have nothing else to identify the entry (the pivot is not present).
-            if (id.getValue().getDatasets().containsKey(SyncReplSourceService.DELETED_ENTRY)
-                    && (task.getDestinationService() instanceof SimpleJndiDstService)) {
-               deleteEntry();
-            } else {
-                IBean data = abstractSynchronize.getBean(task, fromSource ? task.getSourceService() : task.getDestinationService(),
-                        id.getKey(), id.getValue(), true, fromSource);
-                run(data);
-            }
+		try {
```
elecharny, line 46: Does not have any sense when we are dealing with a deleted entry: there is *no* entry to fetch.
Please keep the delete() method separated from the run(Bean).

davidcoutadeur:
After discussion, it appears that the feature can't be implemented in a generic way:

We need the source pivot ids to determine the destination mainIdentifier, but we don't have these pivots ids as they are not present in the syncrepl delete operation, nor in the source entry as it has already been deleted.

So this feature will be closed and not included in lsc-core. The PR and the associated branch can be preserved for people willing to implement a specific patch on their instance.

PR #543, opened 2026-07-23
Retry indefinitively when connection is broken

elecharny:
To be check if we want to add this PR into master.

As it's an infinite retry, the user will have to kill the process to have it stop.


elecharny:
The branch has been renamed to 489-retry

PR #555, opened 2026-08-07
auto register plugins (#553)

### LINKED ISSUE #553: auto-register plugins
Find a way for auto-registering plugins.

This should be done by generating the env var `LSC.PLUGINS.PACKAGEPATH` (note: separated by `:`) in /usr/bin/lsc

Maybe we could auto-discover the class `org.lsc.plugins.connectors.*.generated` in the libs, else we could add a new file for each plugin that reference this class.

davidcoutadeur:
Fixes #553


davidcoutadeur:
Implementation is done and tested

TODO:
* add documentation about the ways to load plugins
* into plugins projects: package the file loading the plugin

PR #558, opened 2026-08-20
Fixed the Java 1.8 compatibility issue (#556]

### LINKED ISSUE #556: Don't use Java code that is not Java 1.8 compatible in LSC
Currently, due to some old libraries we are using (`javax.transaction.xa`), we need to keep the code compatible with Java 1.8, otherwise we have some errors in Eclipse (due to the module being imported twice, something related to Java 9 modules handling).

Some code have been added recently that is not compatible in the `Ldap2LdapHookSyncTest` class:

```
	// Method for sorting arrays in JSON nodes
	private void sortArrays(JsonNode node) {
		if (node.isArray()) {
			for (JsonNode child : node) {
				sortArrays(child);
			}
			// Sort elements in arrayNode
			var arrayNode = (ArrayNode) node;
			var iter = arrayNode.elements();
			var sortedElemsCopy = new ArrayList<JsonNode>(arrayNode.size());
			while(iter.hasNext()) {
				var n = iter.next();
				sortedElemsCopy.add(n);
			}
			sortedElemsCopy.sort(nodeComparator);
...
```

The `var` keyword was introduced in Java 10 and makes the code non compatible with Java 1.8. Beside that, it makes the code difficult to read...

Another option would be to fix the module problem, but it most certainly requires switching to some more recent versions of gthe transaction API.

elecharny:
Fixes #556

### DIFF
--- a/src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
+++ b/src/test/java/org/lsc/Ldap2LdapHookSyncTest.java
@@ -355,11 +355,11 @@ public class Ldap2LdapHookSyncTest extends CommonLdapSyncTest {
 				sortArrays(child);
 			}
 			// Sort elements in arrayNode
-			var arrayNode = (ArrayNode) node;
-			var iter = arrayNode.elements();
-			var sortedElemsCopy = new ArrayList<JsonNode>(arrayNode.size());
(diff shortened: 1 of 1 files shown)


coudot:
Ok for me, you can merge

